PR #17, opened 2025-04-29
First cut at removing the opt-out language

martinthomson:
This was as little as I could manage with respect to turning "opt-out" to something more general: "preferences" or "expressions of preference".

There are a bunch of other challenges to address that we should probably look into.  Some of those change the language a lot more, but I thought that it was a good idea to get rid of the opt-out bias as a first order of business.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -46,19 +46,28 @@ informative:
 
 --- abstract
 
-This document proposes a standardized vocabulary of use cases that can be targeted when expressing machine-readable opt-outs related to Text and Data Mining (TDM) and AI training. The vocabulary is agnostic to specific opt-out mechanisms and enables declaring parties to communicate restrictions or permissions regarding the use of their digital assets in a structured and interoperable manner. It defines three key use cases—TDM, AI Training, and Generative AI Training—which can be referenced by opt-out systems to ensure consistent interpretation across different implementations.
+This document proposes a standardized vocabulary for expressing preferences related to how content is consumed by automated processing systems.
+This vocabulary allows for the creation of structured declarations about restrictions or permissions for use of digital assets.
+The vocabulary is agnostic to the means by which it is conveyed.
+The definitions in the vocabulary enable a shared understanding between entities that express their preference and those that consume the associated digital assets.
 
 --- middle
 
 # Introduction
 
-The purpose of this document is to provide a common vocabulary that can be used for machine-readable opt-outs by parties who wish to restrict the use of their assets for the purpose of AI training and other forms of Text and Data Mining (TDM).
+This document defines a common vocabulary of terms for automated systems that process digital assets.
+The primary purpose of this vocabulary is to enable machine-readable expressions of preferences about how digital assets are used by automated processing systems, such as AI training and other forms of Text and Data Mining (TDM).
 
-The elements of the vocabulary can be used to describe, in a standardized way, the types of uses that a declaring party may wish to restrict (or allow), thereby ensuring that opt-outs can be communicated, processed and stored in a consistent and interoperable manner.
+The elements of the vocabulary can be used to describe, in a standardized way, the types of uses that a declaring party may wish to explicitly restrict or allow.
+This ensures that preferences - either to indicate an opt-out or to explicitly admit a given usage - can be communicated, processed and stored in a consistent and interoperable manner.
 
-The vocabulary is agnostic to the technical implementations of opt-out systems and is designed to ensure that opt-out information can be effectively exchanged between different systems. The vocabulary is intended to govern the use of works in the context of training AI models and other forms of TDM but does not concern itself with the collection of training data (crawling). In particular the vocabulary is not intended for expressing instructions or restrictions related to crawling for the purpose of building a search index, as there are already more specific standards and protocols for this purpose including but not limited to {{!RFC9309}}.
+The vocabulary is agnostic to the technical implementations of how systems process digital assets.
+It is designed to ensure that preference information can be effectively exchanged between different systems.
 
-The vocabulary is intended to both work in contexts where such opt-outs expressed to the declaring party give rise to legal obligation (such as rights reservation made by rightholders) and in contexts where this is not the case. It is without prejudice to applicable laws and the applicability of exceptions and limitations.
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -46,19 +46,28 @@ informative:
 
 --- abstract
 
-This document proposes a standardized vocabulary of use cases that can be targeted when expressing machine-readable opt-outs related to Text and Data Mining (TDM) and AI training. The vocabulary is agnostic to specific opt-out mechanisms and enables declaring parties to communicate restrictions or permissions regarding the use of their digital assets in a structured and interoperable manner. It defines three key use cases—TDM, AI Training, and Generative AI Training—which can be referenced by opt-out systems to ensure consistent interpretation across different implementations.
+This document proposes a standardized vocabulary for expressing preferences related to how content is consumed by automated processing systems.
```
lrosenthol, line 49: Should we bring up the whole gathered vs. processed dichotomy here?

draft-ietf-aipref-vocab.md
```diff
@@ -46,19 +46,28 @@ informative:
...
+This vocabulary allows for the creation of structured declarations about restrictions or permissions for use of digital assets.
```
lrosenthol, line 50:
```suggestion
This vocabulary allows for the creation of declarations about restrictions or permissions for use of digital assets.
```

lrosenthol: structured is a feature of serialization, and this document doesn't cover serialization

draft-ietf-aipref-vocab.md
```diff
@@ -46,19 +46,28 @@ informative:
...
-The purpose of this document is to provide a common vocabulary that can be used for machine-readable opt-outs by parties who wish to restrict the use of their assets for the purpose of AI training and other forms of Text and Data Mining (TDM).
+This document defines a common vocabulary of terms for automated systems that process digital assets.
+The primary purpose of this vocabulary is to enable machine-readable expressions of preferences about how digital assets are used by automated processing systems, such as AI training and other forms of Text and Data Mining (TDM).
 
-The elements of the vocabulary can be used to describe, in a standardized way, the types of uses that a declaring party may wish to restrict (or allow), thereby ensuring that opt-outs can be communicated, processed and stored in a consistent and interoperable manner.
+The elements of the vocabulary can be used to describe, in a standardized way, the types of uses that a declaring party may wish to explicitly restrict or allow.
+This ensures that preferences - either to indicate an opt-out or to explicitly admit a given usage - can be communicated, processed and stored in a consistent and interoperable manner.
```
lrosenthol, line 62:
```suggestion
This ensures that preferences - either to indicate an opt-out or to explicitly admit a given usage - can be communicated and processed in a consistent and interoperable manner.
```

lrosenthol: storage implies serialization

draft-ietf-aipref-vocab.md
```diff
@@ -46,19 +46,28 @@ informative:
...
+The vocabulary is agnostic to the technical implementations of how systems process digital assets.
+It is designed to ensure that preference information can be effectively exchanged between different systems.
 
-The vocabulary is intended to both work in contexts where such opt-outs expressed to the declaring party give rise to legal obligation (such as rights reservation made by rightholders) and in contexts where this is not the case. It is without prejudice to applicable laws and the applicability of exceptions and limitations.
+The vocabulary is intended to govern the use of works in the context of training AI models and other forms of TDM but does not concern itself with the mechanisms involved in obtaining content (that is, crawling).
```
lrosenthol, line 67: I would move this to the top, since this, IMO is a KEY PART of this work!

draft-ietf-aipref-vocab.md
```diff
@@ -67,15 +76,15 @@ The vocabulary is intended to both work in contexts where such opt-outs expresse
 # Definitions
 
 * **Asset:** A digital file or stream of data, usually with associated metadata.
-* **Declaring party:** The entity that expresses an opt-out with regards to an Asset.
+* **Declaring party:** The entity that expresses an preference with regards to an Asset.
```
lrosenthol, line 79:
```suggestion
* **Declaring party:** The entity that expresses a preference with regards to an Asset.
```

lrosenthol (changes requested):

draft-ietf-aipref-vocab.md
Jo-Levy, line 62: Stored can also imply that the machine-readable preference expressed is stored without serialization for later review/consumption by future downstream users. Other ways to express this concept: communicated, processed, and "maintained for future use."

draft-ietf-aipref-vocab.md
martinthomson: I appreciate the edit.  It's better than mine, which was aiming for the minimum change.

I still need to consult with Paul about how we want to proceed with this.  It might be easier if we did a complete rewrite (as you have done) than pick at the edges.  However, that reduces the traceability of our changes.

draft-ietf-aipref-vocab.md
martinthomson: Yet.  My understanding of the defined scope of this document is that it will include serialization and processing.

lrosenthol: Yes and no @martinthomson .

My understanding was that we agreed to start the document with a general purpose vocabulary, and then we would use that in a "second part" which would be a specific serialization/processing use case (e.g., HTTP headers)

martinthomson: The demarcation that I had in mind was that we'd describe a general abstract model here and an exemplary serialization.  The HTTP/robots.txt attachment document would normatively depend on that serialization, but other attachment methods might choose instead to normatively depend on the abstract model instead and define their own serialization.

Now that I spent more words, does that match your understanding?

lrosenthol: Yes, I agree with that...and means I continue to stand by comment about removing "structured".

draft-ietf-aipref-vocab.md
paul2keller: That is indeed what we tried to imply by saying the that preferences can be stored.

martinthomson:
Paul and I have discussed this and we're going to take this and continue to iterate.  There are a few minor points where I think @lrosenthol is more correct, but we'll need to iterate a few more times before we're done.

PR #26, opened 2025-05-15
Define an exemplary format

### LINKED ISSUE #8: Vocabulary interface to preferences
Discussed in the Brussels meeting. The vocabulary (even when mapped to preferences; see #7) does not specify how an attachment mechanism should express it. 

Sense of the room in Brussels was that:
* The vocabulary should define an abstract model
* The vocabulary should also define a preferred / recommended mapping to strings (i.e., a serialisation)
* Individual attachment mechanisms can opt into using that serialisation, or may define their own

martinthomson:
As discussed, we want to define a recommended format.  This executes on that.  It uses Structured Fields, which will ensure that we have a very easy translation to HTTP fields.  It also fits neatly into the other formats we are considering.

There are some interesting implications that are worth highlighting. The text I've added goes into some of that in more detail, but just to list them for visibility:

* This is a sequence of bytes, not a string.  That has implications for how it is included in some formats, but as RFC 9651 limits its syntax to a pretty small subset of the ASCII range, this is mostly just a theoretical consideration, not a practical one.

* Parsing errors in RFC 9651 means that you lose everything.  No preference survives.  That means that authoring errors have consequences.  This is an artifact of the format being more of a computer-to-computer language, rather than being tolerant of human error.  I'm not thrilled with this, but recognize the trade-off involved.

* Repeated keys mean that only the last one is taken.  Again, not thrilled with this, but think that the trade is tolerable.

* Labels are lowercase only.  A lot of the list traffic has started to use "AI=whatever"; RFC 9651 doesn't let us shout.  I think that this is fine, but it's another minor ding.

* Some patterns of use that might be natural are invalid.  For instance, spaces around the "=" is invalid, so "ai = n" would result in a parsing error and no preference being registered.

Closes #8.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -211,11 +211,12 @@ Systems referencing the vocabulary must not introduce additional categories that
 # Exemplary Serialization Format {#format}
 
 This section defines an exemplary serialization format for preferences.
-The format describes how the abstract model could be turned into a Unicode string or sequence of bytes.
+The format describes how the abstract model could be turned into Unicode text or sequence of bytes.
 
 The format relies on the Dictionary type defined in {{Section 3.2 of !FIELDS=RFC9651}}.
 The dictionary keys correspond to usage categories
-and the dictionary values correspond to explicit preferences.
+and the dictionary values correspond to explicit preferences,
+which can be either "y" or "n"; see {{y-or-n}}.
 
 For example, the following is a preference to allow AI training ({{ai}}),
 disallow generative AI training ({{genai}}), and
@@ -243,7 +244,7 @@ lowercase latin characters (a-z), digits (0-9), "_", "-", ".", or "*".
 These are encoded using the mappings in {{!ASCII=RFC0020}}.
 
 
-## Preference Labels
+## Preference Labels {#y-or-n}
 
 The abstract model used has two options for preferences associated with each category:
 allow and disallow.
@@ -332,7 +333,8 @@ ai=y, ai="n", genai=n, genai, tdm=n, tdm=()
 
 If the parsing of the Dictionary fails, no preferences are expressed.
 This includes where keys include uppercase characters,
-as this format is case sensitive.
+as this format is case sensitive
+(more correctly, it operates on bytes, not strings).
 
 
 ## Alternative Formats
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -315,7 +315,7 @@ as this format is case sensitive
 
 ## Alternative Formats
 
-This format is only a exemplary way to represent preferences.
+This format is only an exemplary way to represent preferences.
 The model described in {{model}}, can be used without this serialization.
 
 
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -78,7 +78,7 @@ The vocabulary is intended to work in contexts where such preferences result in 
 * **Asset:** A digital file or stream of data, usually with associated metadata.
 * **Declaring party:** The entity that expresses a preference with regards to an Asset.
 
-# Statements of Preference
+# Statements of Preference {#model}
 
 The vocabulary is a set of categories,
 each of which is defined to cover a class of usage for assets.
@@ -164,19 +164,19 @@ The figure below shows the relationship between these categories:
 
 This list of specific use cases may be expanded in the future, should a consensus emerge between stakeholders, to include categories that address additional use cases as they emerge. In addition to these categories defined in the vocabulary, it is also expected that some systems implementing this vocabulary may extend this list with additional categories for their particular needs.
 
-## Text and Data Mining (TDM) Category
+## Text and Data Mining (TDM) Category {#tdm}
 
 The act of using one or more assets in the context of any automated analytical technique aimed at analyzing text and data in digital form in order to generate information which includes but is not limited to patterns, trends and correlations.
 
 The overarching TDM category is based on the definition of Text and Data Mining in Article 2(2) of {{EUCD2019}}.
 
-## AI Training Category
+## AI Training Category {#ai}
 
 The act of training machine learning models or artificial intelligence (AI).
 
 The use of assets for AI Training is a proper subset of TDM usage.
 
-## Generative AI Training Category
+## Generative AI Training Category {#genai}
 
 The act of training General Purpose AI models that have the capacity to generate text, images or other forms of synthetic content, or the act of training other types of AI models that have the purpose of generating text, images or other forms of synthetic content.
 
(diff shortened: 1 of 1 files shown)


paul2keller:
Thanks @martinthomson for this. This looks good to me (with the caveat, that I am not really in a position to evaluates the merits of the proposal versus other ways of doing this, but it seems fit for purpose). 

One thing I was wondering after reading this, is how the hierarchy between categories is dealt with. based on what we have tdm=n also means ai=n and genai=n without the latter two having to be explicitly declared. wouldn't that be something that also needs to be addressed by the proposed processing algorithm?

martinthomson:
I realized that same thing myself today.  I wasn't sure if I should fix it here or do a follow-up.  I'll take another look tomorrow.

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
 
 The vocabulary does not preclude the use of other specific categories. Any statement of preference based on this vocabulary shall not be interpreted as restricting the use of the work(s) strictly for the purpose of search and discovery as long as no restriction is declared through search-specific means such as {{!RFC9309}}.
 
-## Vocabulary Extensions
+## Vocabulary Extensions {#vocab-extension}
 
 Systems referencing the vocabulary must not introduce additional categories that include existing categories defined in the vocabulary or otherwise include additional hierarchical relationships.
 
 
+# Serialization Format {#format}
```
lrosenthol, line 211:
```suggestion
# Exemplary Serialization Format {#format}
```

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+
+This section defines an exemplary serialization format for preferences.
+The format describes how the abstract model could be turned into a Unicode string or sequence of bytes.
```
lrosenthol, line 214: By Unicode, you really mean UTF-8, yes?  If so, then say that since Unicode can be encoded in many ways.

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+
+The format relies on the Dictionary type defined in {{Section 3.2 of !FIELDS=RFC9651}}.
+The dictionary keys correspond to usage categories
+and the dictionary values correspond to explicit preferences.
+
```
lrosenthol, line 219: Add something here about the values, `y` and `n`, so that it is clear what they represent.  Or just refer to the later section

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+| Category               | Label | Reference |
+|:-----------------------|:------|:----------|
+| Text and Data Mining   | tdm   | {{tdm}}   |
+| AI Training            | ai    | {{ai}}    |
+| Generative AI Training | genai | {{genai}} |
+{: #t-category-labels title="Mappings for Categories"}
+
+Any mapping for a new usage category can only use
+lowercase latin characters (a-z), digits (0-9), "_", "-", ".", or "*".
+These are encoded using the mappings in {{!ASCII=RFC0020}}.
```
lrosenthol, line 243: Why do you say Unicode earlier, but now limit to ASCII?  Which is it?

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+
+| Category               | Label | Reference |
+|:-----------------------|:------|:----------|
+| Text and Data Mining   | tdm   | {{tdm}}   |
+| AI Training            | ai    | {{ai}}    |
+| Generative AI Training | genai | {{genai}} |
+{: #t-category-labels title="Mappings for Categories"}
+
+Any mapping for a new usage category can only use
+lowercase latin characters (a-z), digits (0-9), "_", "-", ".", or "*".
```
lrosenthol, line 242: I'd remove `*`, it has confusing semantics.  

Also, `.` is frequently used to separate "namespaces" - do we want to have that semantic?  If so, should we make that official?

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+These are encoded using the mappings in {{!ASCII=RFC0020}}.
+
+
+## Preference Labels
+
+The abstract model used has two options for preferences associated with each category:
```
lrosenthol, line 248: at this time - unless we add a third in the vocab.

And if we do, how would that be serialized?

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+allow and disallow.
+These are mapped to single byte Tokens ({{Section 3.3.4 of !FIELDS}})
+of "y" and "n", respectively (without the quotes).
```
lrosenthol, line 251:
```suggestion
of `y` and `n`, respectively.
```

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+{{vocab-extension}} addresses the considerations for defining new categories.
+New labels might also be defined for other types of extension
+that do not assign a preference to a usage category.
+In either case, when processing a parsed Dictionary to obtain preferences,
+any unknown labels MUST be ignored.
+
+The Dictionary syntax ({{Section 3.2 of !FIELDS}}) can associate parameters
+with each key-value pair.
+This document does not define any semantics for any parameters that might be included.
+When processing a parsed Dictionary to obtain preferences,
```
lrosenthol, line 286: So you think we should support parameters in the syntax but not provide any guidance on their usage?  Why not just disallow them?  Do we have a use case for them?

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+any unknown parameters MUST be ignored.
+
+
+## Processing Algorithm {#processing}
```
lrosenthol, line 290: Why do we need this here?  Isn't this defined in RFC 9651?  If so, just point there.

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+The parsing algorithm for a Dictionary
+produces a keyed collection of values,
+each with a possibly-empty set of parameters.
+The parsing process guarantees that each key has at most one value and parameters.
+
+To obtain preferences for each of the categories in the vocabulary,
+iterate through the categories.
+For the label that corresponds to that category (see {{t-category-labels}}),
+obtain the corresponding value from the collection,
+disregarding any parameters.
```
lrosenthol, line 305: See - you are ignoring them anyway, so let's disallow from syntax to simplify the parsing rules!

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+A preference is assigned as follows:
+
+* If the value is a Token with a value of "y" (not including quotes),
```
lrosenthol, line 308:
```suggestion
* If the value is a Token with a value of `y`,
```

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+  the associated preference is to allow that category of use.
+
+* If the value is a Token with a value of "n" (not including quotes),
```
lrosenthol, line 311:
```suggestion
* If the value is a Token with a value of `n`,
```

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+This means that duplicating the same key could result in unexpected outcomes.
+For example, the following expresses no preferences:
+
+~~~
+ai=y, ai="n", genai=n, genai, tdm=n, tdm=()
+~~~
+
+If the parsing of the Dictionary fails, no preferences are expressed.
+This includes where keys include uppercase characters,
+as this format is case sensitive.
```
lrosenthol, line 335: If this format is based on bytes and not strings - there is no such thing as case sensitivity, as that only applies to strings with a text encoding...which this does not have!

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+
+
+## Alternative Formats
+
+This format is only a recommended way to represent preferences.
```
lrosenthol, line 340:
```suggestion
This format is only a exemplary way to represent preferences.
```

lrosenthol (dismissed):

draft-ietf-aipref-vocab.md
martinthomson: Hm, Unicode string has a definition that doesn't match what I wanted: https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G32765

What I'm looking for is probably Unicode text: a term without apparent definition, though Unicode uses it often.  Or maybe "coded character sequence": https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G45138  (which is spectacularly opaque, so I might just use "Unicode text").

WDYT?

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
+| AI Training            | ai    | {{ai}}    |
+| Generative AI Training | genai | {{genai}} |
+{: #t-category-labels title="Mappings for Categories"}
+
+Any mapping for a new usage category can only use
+lowercase latin characters (a-z), digits (0-9), "_", "-", ".", or "*".
+These are encoded using the mappings in {{!ASCII=RFC0020}}.
+
+
+## Preference Labels
```
martinthomson, line 246:
```suggestion
## Preference Labels {#y-or-n}
```

draft-ietf-aipref-vocab.md
```diff
@@ -203,11 +203,144 @@ If arrangements, such as contracts exist that explicitly permit the use of that
...
 
 
+# Serialization Format {#format}
+
+This section defines an exemplary serialization format for preferences.
+The format describes how the abstract model could be turned into a Unicode string or sequence of bytes.
+
+The format relies on the Dictionary type defined in {{Section 3.2 of !FIELDS=RFC9651}}.
+The dictionary keys correspond to usage categories
+and the dictionary values correspond to explicit preferences.
```
martinthomson, line 218:
```suggestion
and the dictionary values correspond to explicit preferences,
which can be either "y" or "n"; see {{y-or-n}}.
```

draft-ietf-aipref-vocab.md
martinthomson: Given the range from which the codepoints are drawn, does it matter?

draft-ietf-aipref-vocab.md
martinthomson: The processing in RFC 9651 gets you a dictionary, not our abstract model.

martinthomson (commented):
Thanks for the review @lrosenthol.  This is very helpful.

There are a few things here that I think are worth raising on-list.  I've got a few things to tweak on top of the suggestions as well.

draft-ietf-aipref-vocab.md
martinthomson:
```suggestion
The format describes how the abstract model could be turned into Unicode text or sequence of bytes.
```

draft-ietf-aipref-vocab.md
lrosenthol: Technically no, but it's reads strange to me - which is why I added the comment.

draft-ietf-aipref-vocab.md
lrosenthol: I'll re-read 9651 and see if there are any issues with subsetting.

draft-ietf-aipref-vocab.md
```diff
@@ -308,12 +309,13 @@ ai=y, ai="n", genai=n, genai, tdm=n, tdm=()
...
 This includes where keys include uppercase characters,
-as this format is case sensitive.
+as this format is case sensitive
+(more correctly, it operates on bytes, not strings).
 
 
 ## Alternative Formats
 
-This format is only a recommended way to represent preferences.
+This format is only a exemplary way to represent preferences.
```
lrosenthol, line 318:
```suggestion
This format is only an exemplary way to represent preferences.
```

PR #27, opened 2025-05-15
Remove opt-out from the title

martinthomson:
The changes we made to introductory text didn't catch the title, which now looks very wrong.

Happy to have this wordsmithed.  This is just the best I could come up with.


paul2keller (approved):
Looks good to me. If we end up having both TDM and search as categories in the vocab (which we are aiming for) then this probably will need additional wordsmithing at a later point but for now I think this is good as is and better aligned with our chartered purpose

PR #31, opened 2025-05-21
Shorter intro/abstract

martinthomson:
This takes some of the feedback from the list and makes a few changes.

The main changes are to cut content...

...from the abstract, which is either repeated more clearly in the intro
or in the first sentences.

...from the introduction, with a few small rewordings.

I've also changed sentence structure of some parts, to break up longer
sentences and make them easier to digest, I hope.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -49,7 +49,7 @@ informative:
 This document proposes a standardized vocabulary for expressing preferences related to how digital assets are used by automated processing systems.
 This vocabulary allows for the creation of structured declarations about restrictions or permissions for use of digital assets by such systems.
 The vocabulary is agnostic to the means by which it is conveyed.
-The definitions in the vocabulary facilitate a shared understanding between entities that express such preferences and those that use the associated digital assets.
+The definitions in the vocabulary facilitate a shared understanding between entities that express such preferences and those that might use the associated digital assets.
 
 --- middle
 
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -46,28 +46,41 @@ informative:
 
 --- abstract
 
-This document proposes a standardized vocabulary for expressing preferences related to how digital assets are used by automated processing systems.
-This vocabulary allows for the creation of structured declarations about restrictions or permissions for use of digital assets by such systems.
-The vocabulary is agnostic to the means by which it is conveyed.
-The definitions in the vocabulary facilitate a shared understanding between entities that express such preferences and those that might use the associated digital assets.
+This document proposes a standardized vocabulary for expressing preferences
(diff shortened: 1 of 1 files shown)


paul2keller:
I agree that most edits improve the text, but I think the edits to the last para do not work. I have a number of issues with the new language, but the most important is that where the old language implied that this is intended to work in both contexts the new language says that it is "**intended to work** in contexts where such preferences result in legal obligations" and then continues that "The vocabulary **can be used** in contexts where there is no specific legal protections". To me this somehow implies a hierarchy and I do not think that is correct. 

for now I would suggest reverting to the previous language here (I also do not think that we need to purge the document of all mentions of Text and Data Mining. This is an established legal concept in a number of different jurisdictions that applies to the types of uses for which we develop this vocabulary and we cannot make it go away by simply not mentioning TDM.

paul2keller (changes requested):
I agree that most edits improve the text, but I think the edits to the last para do not work. I have a number of issues with the new language, but the most important is that where the old language implied that this is intended to work in both contexts the new language says that it is "intended to work in contexts where such preferences result in legal obligations" and then continues that "The vocabulary can be used in contexts where there is no specific legal protections". To me this somehow implies a hierarchy and I do not think that is correct.

for now I would suggest reverting to the previous language here (I also do not think that we need to purge the document of all mentions of Text and Data Mining. This is an established legal concept in a number of different jurisdictions that applies to the types of uses for which we develop this vocabulary and we cannot make it go away by simply not mentioning TDM.

draft-ietf-aipref-vocab.md
```diff
@@ -46,28 +46,41 @@ informative:
...
+It is designed to ensure that preference information can be exchanged between different systems
+and consistently understood.
+
+The vocabulary is intended to work in contexts
+where such preferences result in legal obligations.
+This might include rights reservations made by rightholders
+in jurisdictions with conditional exceptions on copyright.
+The vocabulary can be used in contexts where there is no specific legal protections.
+That is, preferences can be made without prejudice to applicable laws
+and the applicability of exceptions and limitations to copyright.
```
TimidRobot, line 83: I find this paragraph a little difficult to parse. It _feels_ like the first sentence contradicts the third.

Maybe something more like:

The vocabulary is intended to work both in contexts where such preferences result in legal obligations and in contexts where there are no specific legal protections. In contexts with resulting legal obligations, this might include rights reservations made by rightholders in jurisdictions with conditional exceptions on copyright. In contexts where there are no specific legal protections, preferences can be made without prejudice to applicable laws and the applicability of exceptions and limitations to copyright.

martinthomson:
That's fair.  I'll open a separate pull request for that paragraph.  @TimidRobot struggled with my alternative - also fair - so it's clearly not there yet.  I do think that the last paragraph is the hardest to read, with very long sentences, but we can work on that separately.

PR #33, opened 2025-05-21
Added inference category

paul2keller:
This is a first attempt at a definition of the inference category. This is taken from the the C2PA's do not train assertion https://opensource.contentauthenticity.org/docs/manifest/assertions-actions#do-not-train-assertion and is based on their definition of `c2pa.ai_inference`. 

I would consider this to be a staring point for discussion and I do not have a strong attachment to the proposed language.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -158,6 +158,9 @@ The act of training General Purpose AI models that have the capacity to generate
 
 The use of assets for Generative AI Training is a proper subset of AI Training usage.
 
+## Inference Category
+
+To act of using one or more assets as input to a trained AI/ML model for the purposes of inferring a result.
 
 # Usage
 


draft-ietf-aipref-vocab.md
```diff
@@ -158,6 +158,9 @@ The act of training General Purpose AI models that have the capacity to generate
 
 The use of assets for Generative AI Training is a proper subset of AI Training usage.
 
+## Inference Category
+
+To act of using one or more assets as input to a trained AI/ML model for the purposes of inferring a result.
```
martinthomson, line 163: I think that the "inferring a result" is a bit of a strange turn of phrase.

```suggestion
The act of using one or more assets as input to a trained AI/ML model
as part of the operation of that model
(as opposed to the training of the model).
```

draft-ietf-aipref-vocab.md
```diff
@@ -158,6 +158,9 @@ The act of training General Purpose AI models that have the capacity to generate
 
 The use of assets for Generative AI Training is a proper subset of AI Training usage.
 
+## Inference Category
```
martinthomson, line 161:
```suggestion
## AI Inference Category {#inference}
```

I think that we need to include AI in the title.  The anchor is just for my benefit later, for #26.

draft-ietf-aipref-vocab.md
paul2keller: This makes sense to me (in terms of substantive change) I just wanted to start any discussion from some existing definition, so i wonder if it makes sense to include your edit now or after we have shared this with the list (here I would apply the same timing as in the Search case, wait until some of the other discussions have died down)

Issue #1, opened 2025-04-10
Search and Discovery


mnot:
Discussed in the Brussels meeting - there was broad agreement in the room that because search and discovery are often intertwined with AI, and because robots.txt also implicates search and discovery, it is important to attempt to address these use cases in our vocabulary.

martinthomson:
Discussed at the design team meeting and this was pretty uncontroversial.  There is some text in #66 that adds some context.

PR #34, opened 2025-05-21
Add Search Category

paul2keller:
Very preliminary attempt of defining a search category in line with #1, that combines the use for building a search engine and the use of assets as input for AI search engines. 

I have relatively low confidence in this approach, so this is really meant as a starting point for a discussion and nothing more


martinthomson (commented):
Some suggestions; let's start a thread on the mailing list.  Perhaps after some of the current discussion settles a little.

paul2keller:
Thanks for the suggestions, which I have approved. I agree with the comment about posting this to the list (which I was planning to do after you had a look and after the current discussions have dried down).

PR #39, opened 2025-06-05
Followup for inference additions

martinthomson:


martinthomson:
I guess I have a very broad interpretation of what TDM is, let's open an issue for that and leave this unresolved until we have that discussion.

PR #43, opened 2025-06-16
Add attachment draft

martinthomson:
This is a direct copy, with just a name change and the necessary metadata updates.


martinthomson:
@mnot @sureshkrishnan you should add @garyillyes to the repo, if you haven't already.

mnot:
Invited.

PR #57, opened 2025-06-25
Rewote the language on embedded preferences and removed FTP

lrosenthol:
I rewrote the section on embedded preferences to make it clear that the issue is the content type and not the actual payload, and gave XMP as an example.

Also removed the reference to FTP, since we're pretty sure it won't work.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -149,24 +149,20 @@ Embedding preferences is expected to be an effective means
 of associating preferences with content,
 because it ensures that metadata is always associated with content.
 
-The main challenge with embedding is that
-a different method is needed for each content type.
-That is,
-a different means of conveying preferences
-needs to be defined for each audio, documents, images, video,
-or other content format.
+Each content type - whether audio, documents, images, video, or other - 
+embeds metadata in different ways, however, in most cases it is possible to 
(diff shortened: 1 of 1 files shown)


martinthomson:
RFC 9309 does mention FTP, but that's probably not worth belaboring.

draft-ietf-aipref-attach.md
```diff
@@ -149,24 +149,20 @@ Embedding preferences is expected to be an effective means
...
 
-The main challenge with embedding is that
-a different method is needed for each content type.
-That is,
-a different means of conveying preferences
-needs to be defined for each audio, documents, images, video,
-or other content format.
+Each content type - whether audio, documents, images, video, or other - 
+embeds metadata in different ways, however, in most cases it is possible to 
+embed the same metadata, just in different places. This is, for example, how XMP (ISO 16684)
```
martinthomson, line 154: We'll need a citation for that, even if it is just normative.  Do you have a link?

martinthomson, line 154: Separately, it occurs that maybe we want to acknowledge the potential for the encoding to differ between formats.

lrosenthol: The official ISO version is at https://www.iso.org/standard/75163.html.  The Adobe version is at https://github.com/adobe/XMP-Toolkit-SDK/blob/main/docs/XMPSpecificationPart1.pdf

PR #62, opened 2025-07-14
Add an Exceptions parameter to the Vocabulary draft

TimidRobot:
Add an `exceptions` parameter to the Vocabulary draft.


martinthomson:
@TimidRobot Thanks for sharing this.  I think that - as we discussed - this is better served by a separate internet draft on the topic.  That will give you a chance to contextualize what you are proposing and allow it to progress independently.

PR #65, opened 2025-07-17
Put "synthetic content" to the fore

martinthomson:
The definition of generative AI training is about the generation of synthetic content.  This removes the distinction between foundational or general purpose models and those that are more specialized, because it focuses on whether the model with have the *capability* to generate content.


lrosenthol (approved):
LGTM

PR #73, opened 2025-07-18
Fixes #2 by proposing a new subcategory for GenAI Inference

### LINKED ISSUE #2: Inference-time preferences
Discussed in the Brussels meeting -- there was broad support for also defining at least one vocabulary term for inference-time use of content by models. More than one term may be necessary, but the sense of the room was that we should keep this as simple as possible at the first instance.

lrosenthol:


martinthomson:
After discussion toward the end of our session, we concluded that we wouldn't include this.  For now at least.

Issue #77, opened 2025-07-18
Add Registry-based Preferences


sposth:
Suggested changes to the heading and the first paragraph of '1.3.  Embedded Preferences':

1) Change Heading

"1.3.  Embedded Preferences"
-->
"1.3. Embedded and Registry-Based Preferences" 

2) Change first paragraph

"This document does not define a means of embedding preferences in content. Embedding preferences is expected to be an effective means of associating preferences with content, because it ensures that metadata is always associated with content."

-->

"This document does not define a means of embedding preferences in content **or associating them via content-derived identifiers such as ISCC (ISO 24138:2024). Both embedded preferences and registry-based approaches using content-derived fingerprints are** expected to be an effective means of associating preferences with content, because *they* ensure that metadata is always associated with content."

(Changes in bold)

@martinthomson

martinthomson:
Personally, I don't agree with the assertion that these are equivalently effective means of attachment.  There's a very big difference between these methods of attachment.  I will raise this on the mailing list, but I personally don't think that this is the right way to manage mention of alternative attachment mechanisms (including registries).

hardie:
I would suggest making this a separate  entry, with text like:

This document does not define a means of associating preferences them via content-derived identifiers such as ISCC (ISO 24138:2024). Registry-based approaches using content-derived fingerprints may be common in certain contexts, and we expect future work to define these means.

sposth:
Thanks @martinthomson and @hardie 

I agree with your comments and would like to suggest a version based on the previous comment that distinguishes the Registry-based Attachment from Embedding.

"1.4. Registry-Based Preferences

This document does not define a means of associating preferences via content-derived identifiers such as ISCC (ISO 24138:2024) and providing them through public registries. Registry-based approaches using content-derived fingerprints may be applicable in certain contexts, particularly where embedding is impractical or unavailable. Such mechanisms may enable persistent association of preferences across distribution channels. We anticipate that future work may define these approaches in more detail."

wumpus:
I am not in favor of mentioning registries at all. That's an extremely different method compared to what we are creating here.

PR #83, opened 2025-07-18
Provide a proposed solution for #77 based on the text in the issue

lrosenthol:
This is for @sposth as a formal PR for his proposal in issue #77 .

I neither support nor reject this proposal - I am simply acting as a facilitator of the discussion.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -172,6 +172,10 @@ Future work might define how preferences might be indicated
 for alternative content distribution or acquisition methods,
 such as email.
 
+## Registry-Based Preferences
+
+This document does not define a means of using unique identifiers and a registry for associating preferences. Registry-based approaches might be applicable in certain contexts, particularly where embedding is impractical or unavailable. Additionally, a registry might enable persistent association of preferences across distribution channels.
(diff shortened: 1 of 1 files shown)


paul2keller (approved):
I am supportive of adding this language, but I think the ultimate decisions should be with the editors of the attachment draft @garyillyes and @martinthomson

draft-ietf-aipref-attach.md
```diff
@@ -172,6 +172,10 @@ Future work might define how preferences might be indicated
 for alternative content distribution or acquisition methods,
 such as email.
 
+## Registry-Based Preferences
+
+This document does not define a means of using unique identifiers and a registry for associating preferences. Registry-based approaches might be applicable in certain contexts, particularly where embedding is impractical or unavailable. Additionally, a registry might enable persistent association of preferences across distribution channels.
```
martinthomson, line 177:
```suggestion
This document does not define a means of using unique identifiers and a registry
for associating preferences.
Registry-based approaches might be applicable in certain contexts,
particularly where embedding is impractical or unavailable.
Additionally, a registry might enable persistent association of preferences
across distribution channels.
```

PR #93, opened 2025-07-21
Provide rough definitions for AI and Training

martinthomson:
These aren't necessarily perfect, but they should help ensure that the definitions that depend on this are clear.


coopdanger:
There are a lot of definitions of AI floating around, does this come from one of them?

PR #96, opened 2025-07-21
robots.txt attachment: make path segment mandatory

bnewbold:
This updates the draft text to make the path segment mandatory instead of optional. The motivation is to make the `Content-Usage` directive read more clearly like the `Allow`/`Disallow` directives, and generally make robots.txt files more legible and understandable to humans (whether they have prior experience with robots.txt or not). This should have almost no impact on actual behavior or what preferences can be expressed. The trade-off is that every line will have two additional bytes for the case of "all paths", and that seems like a reasonable trade off for legibility.

A secondary benefit is that implementation becomes marginally easier: code will know there is always a leading path segment. While optional fields are common in protocols, it is more common to have trailing fields be optional, not leading fields. The current text *is* unambiguous about how to detect whether there is a path segment or not (if there is a leading `/` character), but this makes things just a bit easier to implement correctly.

I think this would be an improvement on the syntax and it would be great to get it in. However, I do want to prioritize moving work forward, and think this improvement is not worth holding things up.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -137,7 +137,7 @@ a server might include the same directive in its "robots.txt" file:
 ~~~
 User-Agent: *
 Allow: /
-Content-Usage: train-ai=n
+Content-Usage: / train-ai=n
 ~~~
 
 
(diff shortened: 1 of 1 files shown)


mnot:
Closing this as it seems to be overcome by events. If you still think there's an underlying issue to address, please open one during WGLC. Thanks.

PR #108, opened 2025-08-01
A more verbose search definition

martinthomson:
This keeps the short definition, but it expands it considerably to cover what seems to be the emerging perspective on the list.

This aims to do three things:

1. Narrow the definition to cover those cases that drive traffic to the original location of assets.

2. Ensure that this allows the use of AI in search applications, so that search preferences override AI use preferences.

3. To ensure that AI overviews, which act as a substitute for visiting the site, are excluded.

Obviously, there will continue to be things on the line.  The draft has text addressing the question of ambiguity.  As has been noted, the law has far better tools for managing that than technical specifications... if there is a dispute, that is.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -264,7 +264,7 @@ Using one or more assets in a search application
 that directs users to the location
 from which the assets were retrieved.
 
-Search applications can be complex 
+Search applications can be complex
 and may serve multiple purposes.
 Only those parts of applications that direct users to the location of an asset
 are included in this category of use.


draft-ietf-aipref-vocab.md
```diff
@@ -260,13 +260,27 @@ The use of assets for AI Use is a proper subset of Automated Processing usage.
...
+Search applications can be complex applications
+that serve multiple purposes.
+Only those parts of applications that direct users to the location of an asset
+are covered by preferences for this category of use.
+Preferences for the Search category apply to those parts of applications
+that provide search capabilities,
+regardless of what other preferences are expressed.
+This includes the use of titles or excerpts from assets
+that are used to help users select between multiple candidate options.
+
```
martinthomson, line 276:
```suggestion
This includes the use of titles or excerpts from assets
that are used to help users select between multiple candidate options.

Preferences for the Search category apply to those parts of applications
that provide search capabilities,
regardless of what other preferences are expressed.
```

draft-ietf-aipref-vocab.md
```diff
@@ -260,13 +260,27 @@ The use of assets for AI Use is a proper subset of Automated Processing usage.
...
-category is to allow preferences to be expressed about search applications,
-independent of other categories of use.
-A distinct Search category allows for preferences specific to search applications,
-even if the use of AI is involved in their implementation.
+Using one or more assets in a search application
+that directs users to the location
+from which the assets were retrieved.
+
+Search applications can be complex applications
+that serve multiple purposes.
```
mnot, line 268:
```suggestion
Search applications can be complex and serve multiple purposes.
```

draft-ietf-aipref-vocab.md
```diff
@@ -260,13 +260,27 @@ The use of assets for AI Use is a proper subset of Automated Processing usage.
...
+Only those parts of applications that direct users to the location of an asset
+are covered by preferences for this category of use.
```
mnot, line 270:
```suggestion
Only those parts of applications that direct users to the location of an asset
are included in this category of use.
```

PR #109, opened 2025-08-06
Expand on the preference overriding text

### LINKED ISSUE #99: Add a more complete list of preference override reasons
Let's collect a list of reasons here and provide them, even if that is not exhaustive.

martinthomson:
This is a light reworking of a proposal from Lila Bailey on the list.

The value of this is that it is far more comprehensive than the previous text, which was directionally helpful, but not thorough enough to give the right impression.  Having more text is often something to avoid, but in this case it seems like it is adding a lot of value in terms of making the *preferences* thing far clearer.

My only concern here -- which is a very minor one -- is that this looks almost exclusively at the overriding of preferences to disallow.  I couldn't come up with a convincing example of a decision to ignore express permission.

Closes #99.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -164,8 +164,6 @@ and those documents that define concrete uses of this vocabulary --
 describe how usage preferences are associated with assets.
 Conformance to the specification means following the normative language
 that defines the construction and interpretation of usage preferences.
-The process of obtaining preferences has very limited scope
-for variation between implementations.
 
 Variation in what usage preferences might be obtained are limited to:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -176,26 +174,59 @@ Variation in what usage preferences might be obtained are limited to:
...
-so that they might access automated captions or summaries.
-Another case might involve the use of assets for research.
-Such overrides might be explicitly permitted in law,
-or could be based on the judgment of individual system users.
+Priorities that could justify overriding preferences include --
+but are not limited to --
+free expression, safety, education, scholarship, research,
+preservation, interoperability, and accessibility.
+
+A choice to override a preference could be
```
mnot, line 203:
```suggestion
A choice to ignore a preference could be
```

martinthomson:
The arrow of time is ever forward; feedback suggests that this is an improvement, even if this isn't necessarily a final state for this text.

PR #115, opened 2025-08-14
Use consistent phrasing in Definitions section

chrisn:
I didn't have chance to comment on the PR before it was merged, sorry. I think it would be good if the definitions followed a more consistent phrasing.


martinthomson (approved):
Thanks for catching my error Chris.

PR #117, opened 2025-08-25
A suggestion for 3.2

mnot:


draft-ietf-aipref-vocab.md
```diff
@@ -165,57 +165,21 @@ The process of obtaining preferences has very limited scope
...
+
+Other mechanisms
+-- technical, legal, or otherwise --
+might enforce stated preferences.
+As such they will determine the consequences for ignoring a stated preference.
+Preferences do not imply control, either in the positive or negative.
+For example, a stated preference allowing a use does not 
+(on its own) imply an unconditional right to use it for that purpose;
+likewise, a stated preference disallowing a use does not
+(on its own) mean that the content can never be used for that purpose.
```
mnot, line 182:
```suggestion
## Enforcement

This specification does not provide any enforcement mechanism,
and conformance to it does not encompass whether
preferences are actually respected during data processing.

Other mechanisms
-- technical, legal, or otherwise --
might enforce stated preferences.
As such they will determine the consequences for ignoring a stated preference.
Preferences do not imply control, either in the positive or negative.
For example, a stated preference allowing a use does not
(on its own) imply an unconditional right to use it for that purpose;
likewise, a stated preference disallowing a use does not
(on its own) mean that the content can never be used for that purpose.
```

paul2keller:
I think we are at risk of loosing substantial safeguards here.We need to have explicit language that there will be situations in which legal regimes override any preferences (the most concrete example, although not the best because Eurocentrism concerns) is the TDM exception for scientific research in Article 3 the EU copyright directive. 

I think the approach to combine (or at least move closer to each other) the legal section and this one makes a lot of sense. And as a reminder there is one other section in the doc that also deals with this issue, that is the 2nd para of the `#introduction` section:

> The vocabulary is intended to be used in jurisdictions where expressing preferences results in legal obligations, as well as where there are no associated legal obligations. In either case, expressing preferences is without prejudice to applicable laws, including the applicability of exceptions and limitations to copyright.

draft-ietf-aipref-vocab.md
```diff
@@ -165,57 +165,21 @@ The process of obtaining preferences has very limited scope
...
+and conformance to it does not encompass whether
+preferences are actually respected during data processing.
+
+Other mechanisms
+-- technical, legal, or otherwise --
+might enforce stated preferences.
+As such they will determine the consequences for ignoring a stated preference.
+Preferences do not imply control, either in the positive or negative.
+For example, a stated preference allowing a use does not
+(on its own) imply an unconditional right to use it for that purpose;
```
martinthomson, line 180:
```suggestion
(on its own) imply a right to use it for that purpose;
```

draft-ietf-aipref-vocab.md
```diff
@@ -165,57 +165,21 @@ The process of obtaining preferences has very limited scope
...
+likewise, a stated preference disallowing a use does not
+(on its own) mean that the content can never be used for that purpose.
```
martinthomson, line 182:
```suggestion
(on its own) mean that the content cannot be used for that purpose.
```

draft-ietf-aipref-vocab.md
```diff
@@ -165,57 +165,21 @@ The process of obtaining preferences has very limited scope
...
+## Enforcement
+
+This specification does not provide any enforcement mechanism,
+and conformance to it does not encompass whether
+preferences are actually respected during data processing.
+
+Other mechanisms
+-- technical, legal, or otherwise --
+might enforce stated preferences.
+As such they will determine the consequences for ignoring a stated preference.
```
martinthomson, line 177:
```suggestion
As such they will determine the consequences for following or not following a stated preference.
```

martinthomson (commented):
Jo Levy mentioned some tweaks that alter the language here.  I'm merely relaying that.  Paul's concerns are more substantive here.

mnot:
I've moved applicability/legal effect up; it likely needs some editorial adjustment.

draft-ietf-aipref-vocab.md
```diff
@@ -165,56 +165,61 @@ The process of obtaining preferences has very limited scope
...
 
-The following lists examples of cases
-where other priorities could override specific preferences:
+The expression of preferences might activate regulatory or legal consequences,
+which has implications for entities that consume those preferences.
+Their interpretation of the meaning of different terms
+could have legal ramifications.
+Different jurisdictions could reach subtly different conclusions
+about the applicability of each category of use
+to specific applications.
```
martinthomson, line 204: This is probably most redundant, but there is still a useful point: how a category applies to a given application could be a judgment call and that might vary between jurisdictions (if there is a legal effect).

martinthomson (commented):
So there is definitely some overlap here with these two sections.

I think that this could be one section.  Though there are a few things being said here:

1. These are just preferences.
2. However, a statement of preferences can activate laws depending on context.
3. The categories are simple, broad, and therefore will have ambiguities as they are applied to real applications.
4. People who use assets need to take responsibility for that use.
5. (what you removed) There are cases where preferences might not be followed.

The text you added only really covers points 1 and 2.

mnot:
Closing in favour of #118.

PR #119, opened 2025-08-27
Combine Legal and Conformance section

paul2keller:
Another attempt to combine the legal effects and conformance sections that stays closer to the language of the existing conformance section than #118

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -166,18 +166,18 @@ for variation between implementations.
 
 ## Applicability and Effect {#applicability}
 
-This specification provides a set of definitions for different 
-categories of use, plus a system for associating simple 
+This specification provides a set of definitions for different
+categories of use, plus a system for associating simple
 preferences to each (allow, disallow, or no preference; see {{model}}).
 
-This specification does not provide any enforcement mechanism 
-for those preferences, and conformance to it does not encompass 
+This specification does not provide any enforcement mechanism
+for those preferences, and conformance to it does not encompass
 whether preferences are actually respected during data processing.
 
-Preferences do not themselves create rights or prohibitions, 
-either in the positive or the negative. Other mechanisms—technical, 
-legal, contractual, or otherwise—might enforce stated preferences 
-and thereby determine the consequences of following or not following 
+Preferences do not themselves create rights or prohibitions,
+either in the positive or the negative. Other mechanisms—technical,
+legal, contractual, or otherwise—might enforce stated preferences
+and thereby determine the consequences of following or not following
 a stated preference.
 
 An entity that receives usage preferences MAY choose to respect
@@ -187,12 +187,12 @@ how that usage corresponds to the usage categories
 where preferences have been expressed,
 and the applicable legal context.
 
-Usage preferences can be overridden through express agreements 
-between relevant parties, by explicit provisions of law, or 
-through the exercise of discretion in situations where other 
-priorities justify doing so. Priorities that could justify 
-ignoring preferences include—but are not limited to—free 
-expression, safety, education, scholarship, research, 
+Usage preferences can be overridden through express agreements
+between relevant parties, by explicit provisions of law, or
+through the exercise of discretion in situations where other
+priorities justify doing so. Priorities that could justify
+ignoring preferences include—but are not limited to—free
+expression, safety, education, scholarship, research,
 preservation, interoperability, and accessibility.
 
 The following lists examples of cases
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -181,12 +181,11 @@ and thereby determine the consequences of following or not following
 a stated preference.
 
 A data processor MAY choose to respect preferences that it has 
-discovered, according to:
-
-*	an understanding of the nature of the processing being performed 
-and how it corresponds to the usage categories where preferences 
-have been expressed, and
-* the applicable legal context.
+discovered, according to
+an understanding of how the asset is used,
+how that usage corresponds to the usage categories
+where preferences have been expressed,
+and the applicable legal context.
 
 Usage preferences can be overridden through express agreements 
 between relevant parties, by explicit provisions of law, or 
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -216,6 +220,7 @@ where other priorities could override specific preferences:
   in order to develop or use tools that detect harmful content
   according to established terms of use.
 
+Because enforcement is not provided by this specification, the consequences of ignoring preferences may vary depending upon how a given legal jurisdiction recognizes preferences, including how it maps categories of use to specific applications.
```
martinthomson, line 223:
```suggestion
Because enforcement is not provided by this specification,
the consequences of ignoring preferences may vary
depending upon how a given legal jurisdiction recognizes preferences;
see {{applicability}}.
```

draft-ietf-aipref-vocab.md
```diff
@@ -164,34 +164,38 @@ if it correctly follows all normative requirements that apply to it.
...
-
-Usage preferences can be overridden through express agreements
-between relevant parties.
-
-There are also many situations where other priorities could override
-any usage preferences.
-Priorities that could justify ignoring preferences include --
-but are not limited to --
-free expression, safety, education, scholarship, research,
+## Applicability and Effect
```
martinthomson, line 167:
```suggestion
## Applicability and Effect {#applicability}
```

draft-ietf-aipref-vocab.md
```diff
@@ -164,34 +164,38 @@ if it correctly follows all normative requirements that apply to it.
...
+and thereby determine the consequences of following or not following 
+a stated preference.
+
+A data processor MAY choose to respect preferences that it has 
+discovered, according to:
+
+*	an understanding of the nature of the processing being performed 
+and how it corresponds to the usage categories where preferences 
+have been expressed, and
+* the applicable legal context.
```
martinthomson, line 189: A list of two might just be better as a sentence:

```suggestion
discovered, according to
an understanding of how the asset is used,
how that usage corresponds to the usage categories
where preferences have been expressed,
and the applicable legal context.
```

martinthomson (commented):
This is turning out to be a difficult needle to successfully thread.  As with other genuine attempts to improve the text, this is helpful.  I just don't know what razor to apply to decide which is best.

paul2keller:
> This is turning out to be a difficult needle to successfully thread. As with other genuine attempts to improve the text, this is helpful. I just don't know what razor to apply to decide which is best.

yes this is tricky. for me there are two aspects where I would consider this to preserve important elements of the original language we had before attempts to merge the two sections: 

1. the very unambiguous `MAY` in "A data processor MAY choose to respect preferences that is has discovered" that for me is a very important signal that we are talking about "preferences" that in itself do not have legal effect, and 
2. the "could ignore a preference" wording in the examples, that in #118 in some of the examples has changed into "might be allowed by a legal system" which implies that absent an explicit legal authorization such the preference must be followed (at which point is ceases to be a "mere" preference). 

but you are right that these are all subjective choices and there is formal way of establishing which approach is better.

draft-ietf-aipref-vocab.md
```diff
@@ -164,34 +164,37 @@ if it correctly follows all normative requirements that apply to it.
...
+ignoring preferences include—but are not limited to—free 
+expression, safety, education, scholarship, research, 
 preservation, interoperability, and accessibility.
 
-A choice to ignore a preference could be
-explicitly permitted in law
-or be based on the judgement of particular individuals or organizations.
-
 The following lists examples of cases
 where other priorities could override specific preferences:
```
martinthomson, line 199: I think that I prefer Mark's take on this list of things, both the framing and the "might be allowed by a legal system to ignore a preference"

paul2keller: you mean replacing lines 198 and 199 with: 

> For example, a legal system might require adherence to stated preferences,
but allow them to be ignored for specified purposes, such as:

felixreda: I prefer @paul2keller s proposed language insofar as it doesn't refer to a legal system that allows ignoring a certain preference. By default, anything that is not expressly prohibited by a legal system is allowed. Therefore, a legal system rarely is explicit in what it allows. Consequently, "a legal system might ... allow them to be ignored" could create misunderstandings where people might read this as requiring an explicit legal provision that allows somebody to ignore a stated preference. In reality, unless a legal system explicitly requires somebody to _respect_ a stated preference, ignoring it is allowed.

felixreda:
> 1. the very unambiguous `MAY` in "A data processor MAY choose to respect preferences that is has discovered" that for me is a very important signal that we are talking about "preferences" that in itself do not have legal effect, and
> 2. the "could ignore a preference" wording in the examples, that in [Another 3.2 suggestion #118](https://github.com/ietf-wg-aipref/drafts/pull/118) in some of the examples has changed into "might be allowed by a legal system" which implies that absent an explicit legal authorization such the preference must be followed (at which point is ceases to be a "mere" preference).

I had exactly the same reactions to reading the other propsal, so I think this is a very valuable rewrite.

draft-ietf-aipref-vocab.md
```diff
@@ -216,6 +219,10 @@ where other priorities could override specific preferences:
   in order to develop or use tools that detect harmful content
   according to established terms of use.
 
+Because enforcement is not provided by this specification,
+the consequences of ignoring preferences may vary
+depending upon how a given legal jurisdiction recognizes preferences;
+see {{applicability}}.
```
martinthomson, line 225:
```suggestion
depending upon how a given legal jurisdiction recognizes preferences.
```

draft-ietf-aipref-vocab.md
```diff
@@ -164,34 +164,37 @@ if it correctly follows all normative requirements that apply to it.
...
+whether preferences are actually respected during data processing.
+
+Preferences do not themselves create rights or prohibitions, 
+either in the positive or the negative. Other mechanisms—technical, 
+legal, contractual, or otherwise—might enforce stated preferences 
+and thereby determine the consequences of following or not following 
+a stated preference.
+
+A data processor MAY choose to respect preferences that it has 
+discovered, according to
```
martinthomson, line 184:
```suggestion
An entity that receives usage preferences MAY choose to respect
those preferences it has discovered, according to
```

PR #120, opened 2025-09-01
Move vocabulary extensions

martinthomson:
Reword the statement; say it twice for clarity.

I considered adding more text here, even wrote out an example.  The example was convoluted and didn't help.  It even pointed to a way to manage the problem; an ugly way, but a way.  That was even more confusing.  A simple rule is best.


paul2keller (approved):
Agree that this is better (and that moving it up in the document makes more sense)

PR #133, opened 2025-09-02
Response header

mnot:
It's necessary to distinguish it.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -87,7 +87,7 @@ by an automated system.
 This document describes two mechanisms
 for associating preferences with content:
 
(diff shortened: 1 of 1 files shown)


martinthomson (changes requested):
This does not need to be strictly limited to be a response header field.

Consider a PUT.  That could use a *request* header field to signal to the resource that it wants to attach specific usage preferences to the resulting resource.  Maybe it's an issue that this doesn't explicitly mention requests, but I didn't imagine that this was needed.

mnot:
Fair point. The header should mention that it's representation metadata:
  https://httpwg.org/specs/rfc9110.html#representation.data.and.metadata

PR #139, opened 2025-09-03
The field is representation metadata

martinthomson:
I thought that this was clear; it could have been clearer.

Closes #133.


martinthomson:
@mnot, do you think I need to include a PUT example?  It seems like it might be an attractive nuisance to get much deeper into explaining HTTP here.

PR #140, opened 2025-09-03
Rename 'all' to 'bots'

### LINKED ISSUE #97: Short name of top-level group
"all" is probably a bit *too* expansive.  Let's find a better one.

martinthomson:
I was careful to avoid using "all" in the attachment draft, so this only affects the one document.

Closes #97.  At least for now.


paul2keller (approved):
I don't think this really nails it either. the whole vocabulary is about use of assets and not about crawling. I feel the term `bots` may be to closely associated with crawling and that this will cause confusion about what the vocab governs. that being said `bots` is better than `all` so lets go with this for now to see how people react.

PR #142, opened 2025-09-03
Parameters carry through

### LINKED ISSUE #129: Parameters
6.5 says:

>  The parsing process guarantees that each key has at most one value and parameters.

but then:

> obtain the corresponding value from the collection, disregarding any parameters.

So are parameters passed through the algorithm or not?

martinthomson:
This is, of course, useless in the current state of the design.  But we do allow for the possibility that parameters will be defined.

The only design choice here is whether to respect parameters that are attached to invalid preferences.  The answer is that we do that, because maybe that will turn out to be useful and there is no reason obvious reason not to.

Closes #129.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -477,7 +477,8 @@ To process a series of bytes to recover the expressed preferences,
 those bytes are parsed into a Dictionary ({{Section 4.2.2 of FIELDS}}),
 then preferences are assigned to each usage category in the vocabulary.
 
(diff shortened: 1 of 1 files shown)


paul2keller (commented):
I feel this needs to be reviewed by someone with more experience in drafting such specifications. probably @mnot who raised this

mnot:
LGTM. I got slightly tripped up on 

~~~
train-ai;has;parameters="?";
~~~

... because that assigns boolean no to train-ai, but that's ok -- it isn't expressing a preference.

PR #144, opened 2025-09-03
Restructure "other mechanisms" section

### LINKED ISSUE #138: Move other, undefined attachment mechanisms into an appendix?
1.3 and 1.4 seem very tangental, and more commentary about what this document _isn't_ doing.

Move them into an appendix, perhaps?

martinthomson:
Another heading level seems worthwhile.  If only so that we can say that this is intended to be complementary to other mechanisms.

A slight rewording of the registry text was necessary to avoid repetition.

Closes #138.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -145,7 +145,15 @@ Content-Usage: train-ai=n
 This document provides two general purpose methods
 for associating statements of preference with assets
 that are transferred using HTTP.
-This is intended to be complementary with other mechanisms.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-attach.md
```diff
@@ -165,10 +173,12 @@ The mechanisms in this document can be applied to any content type, provided tha
 for alternative content distribution or acquisition methods,
```
paul2keller, line 173: (not my draft, but after reading trough this section it feels to me that line 171-174 logically do not belong in the Embedded preferences section, but should be moved after the Registry-based preferences section.

PR #145, opened 2025-09-03
More clarity about vocabulary extensions

### LINKED ISSUE #122: Extending the vocabulary
> Systems referencing the vocabulary MUST NOT introduce additional categories that include existing categories defined in the vocabulary. That is, new categories of use can be defined as a subset of an existing category, but not a superset

The first sentence implies that "systems referencing the vocabulary" can introduce additional categories. However, 6.4 says

> New labels might be defined to correspond to new usage categories. [Section 4.5](https://ietf-wg-aipref.github.io/drafts/draft-ietf-aipref-vocab.html#vocab-extension) addresses the considerations for defining new categories. [...] [N]ew extensions need to be defined in an RFC that updates this document.

These statements are seemingly in conflict. It might be good to more explicitly discuss the conditions under which new categories can be added in 4.5, and adjust the language used there.

martinthomson:
Be clearer about the need for a RFC-based revision to extend the vocabulary, what the constraints on that will still be.

Then make clear that alternative formats could extend their own use of this vocabulary and that we have already dealt with that later in the document.

Closes #122.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -323,8 +323,10 @@ that include existing categories defined in the vocabulary.
 That is, new categories of use can be defined as a subset of an existing category,
 but not a superset.
 
-Systems that use this vocabulary might define their own extensions,
-see {{mapping}} for how those are managed.
+Systems that use this vocabulary might define their own extensions
+as part of a larger data model.
+{{mapping}} describes how concepts from an alternative format
+might be mapped to this vocabulary.
 
 
 # Applying Statements of Preference {#usage}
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -315,11 +315,17 @@ The use of assets for Search is a proper subset of Automated Processing usage.
...
+Extensions to this vocabulary need to be defined in an RFC
+that updates this document.
+
+Any future extensions to this vocabulary MUST NOT introduce additional categories
 that include existing categories defined in the vocabulary.
 That is, new categories of use can be defined as a subset of an existing category,
 but not a superset.
 
+Systems that use this vocabulary might define their own extensions,
+see {{mapping}} for how those are managed.
```
mnot, line 327: Hmm. That refers to an alternative serialisation, which isn't really what this is about -- it's about an external consumer of the vocabulary that repurposes or augments it outside the scope of this specification. Something can do that without using the alternative serialisation (which is more about attachment than the vocabulary).

PR #175, opened 2025-10-01
Add some text to security

martinthomson:
Right now, this isn't going to be much, because we don't have the format text merged.  When that arrives, we'll need to add some more.

This text is almost applicability-statement material, but I thought it worth putting it here, because it will ensure that attachment methods include security considerations of their own that address the question of authentication.


martinthomson:
This was an old change, which probably needs to be moved to the attachment draft instead.

PR #176, opened 2025-10-01
AI Output and Search categories

martinthomson:
This is the output of the informali, impromptu design team that was formed during the last day of meetings.  This attempts to synthesize ideas from numerous sources, including various email threads on the mailing list, drafts shared (in particular
draft-madhavan-aipref-displaybasedpref) and some ideas that were floating around.

This concentrates on system outputs and aims to primarily address concerns about substitutive use.  That is, the potential for AI to generate content that is different from source material, but which might replace it.

We did not include some of the concepts that were in draft-madhavan-aipref-displaybasedpref about limiting output size and so forth, but recognize that those might make good extensions at a later time.  Those will depend on specific formats (text lengths for text, snippet duration for audio or video, etc..) and we concluded that we want to concentrate on the core concept.

This is a less flexible version of the thing that was presented.  Rather than having separate conditions of use attached to a permission grant, one for linking back and another for requiring direct quoting rather than paraphrasing or summarizing.  These too might need further refinement in future.  For example, we talked about how certain formats might be able to identify which content (text, time, whatever) is OK to quote and which is not.

This does not attempt to address the new problem this raises, which is that indicating that you would like these new categories to be allowed in practice also means that you need to say the same for AI training. That's not a problem this is attempting to solve.  We'll need to do that in some other way.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -313,7 +313,7 @@ The use of assets for AI Output is a proper subset of Automated Processing usage
 
 ## Search {#search}
 
-The Search category of use is a refinement of the AI Output usage,
+The Search category of use is a refinement of the AI Output category,
 with the addition of the following two conditions:
 
 * A reference to the location that the asset was obtained


draft-ietf-aipref-vocab.md
```diff
@@ -426,6 +439,7 @@ Each usage category in the vocabulary ({{vocab}}) is mapped to a short textual l
 | Automated Processing   | bots        | {{bots}}        |
 | AI Training            | train-ai    | {{train-ai}}    |
 | Generative AI Training | train-genai | {{train-genai}} |
+| Search                 | ai-output   | {{ai-output}}   |
```
paul2keller, line 442: I think this should be AI Output

martinthomson:
https://ietf-wg-aipref.github.io/drafts/ai-output/draft-ietf-aipref-vocab.html is a preview.

draft-ietf-aipref-vocab.md
```diff
@@ -289,28 +289,41 @@ The act of training general purpose AI models that have the capacity to generate
...
+and outputs that are presented to automated clients.
 
-Parts of applications that do not direct users to the location of assets,
-such as summaries,
-are not covered by this category of use.
+The use of assets for AI Output is a proper subset of Automated Processing usage.
 
-The use of assets for Search is a proper subset of Automated Processing usage.
+
+## Search {#search}
```
martinthomson, line 310: There is contention about the choice of name here.  The label was chosen in an attempt to align with expectations, but the name should match the definition.

PR #184, opened 2025-11-03
indexing and derivative works

mallory:
Updated sections on AI Output and Search categories to clarify derivative works and indexing requirements as per my comments on the mic today.


martinthomson:
A lot of things have changed since and this affects text that has new proposals that have been extensively discussed.  Closing now as overtaken by events.

PR #186, opened 2025-11-18
Remove Automated Processing and AI Output

martinthomson:
This is about as small a change as I could manage.  With the removal of any concrete examples regarding subsets and supersets, some of the text had to be updated to be more generic.

If we commit to something that doesn't have groupings like this, we'll want to do that.  Because the state of things is still in flux, and because it is less disruptive, I'm going to recommend that we stick with the remnants of nesting for now.


paul2keller (commented):
@martinthomson there is still a reference to the old automated processing category in line 556/497:

train-ai=y, train-ai="n", search=n, search, bots=n, bots=()

PR #191, opened 2026-01-27
Define "foundation model" (align with ISO/IEC 22989)

lexlizx2:
This PR adds an explicit definition of “foundation model” to the AIPREF vocabulary for clarity and consistency.

The definition is aligned with ISO/IEC 22989 (AI — Concepts and terminology), which defines a foundation model as an AI model that can be used for or readily adapted to a wide range of tasks in one or more domains.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -243,10 +243,15 @@ This section defines the categories of use in the vocabulary.
 
 ## Foundation Model Production Category {#train-ai}
 
+
+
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
I like the definition, but I'll need to tweak the formatting.

The paragraph following your change contains more definition, some of which might be redundant with what was added.  Did you consider that?

martinthomson:
After our recent interim, we're going with something else in #207.

Issue #196, opened 2026-04-07
Proposed revision of search category to focus on "substantial" generative use


erinsimon:
### Confirmation

- [x] I have reviewed our [existing issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

There has been substantial debate about how to define a search category that can be used to express a preference for use in traditional forms of search results without including the newer more [narrative overviews ](https://github.com/ietf-wg-aipref/drafts/issues/102)that have been introduced in the last few years.  

Articulating this distinction is [tricky](https://github.com/ietf-wg-aipref/drafts/issues/98) for a few reasons: 
- Traditional search results are the [result of AI processes](https://github.com/ietf-wg-aipref/drafts/issues/173), so the category cannot disclaim all AI use or the input of content to AI systems.
- Traditional search results may involve some amount of [non-verbatim copying](https://github.com/ietf-wg-aipref/drafts/issues/181), such as eliding or mildly rewording text to better highlight relevant aspects.
- The category is not intended to prohibit categorization and labeling of traditional search results, but an overly strict prohibition on generated content could cover the generation of very short category labels.
- Referring instead to "summaries" does not provide sufficient clarity about how much or what kind of generated output is permissible.  Referring instead to the [potential to substitute for the asset](https://github.com/ietf-wg-aipref/drafts/issues/150) begs legal questions as well as introduces ambiguity in scope.
- Requiring the asset to be linked to in the output presumes that the asset will be relevant to the output, and could be interpreted to mean that the output cannot be evaluated for inclusion without a guarantee that the asset is included. 

What is needed is a definition that focuses on the core distinction of the use of the asset as an input to the generation of more than very short amounts of content, but allows for some flexibility in quoting, categorizing, and labeling.  This proposal describes that as "substantial" amounts of "new" content, and as an initial proposal defines substantial to mean sentence-level text.  A numerical threshold could be used instead, but it would have to account for the differences in semantic density between languages.

It is also necessary to harmonize this category with the beneficial uses mentioned in the preamble and the working group discussions, to ensure that it does not impede the accessibility of content through AI-mediated means.  Thus the category clarifies that transforming the modality of the content (e.g., text to speech, translation) is not in scope of the category preference.

This is, of course, a starting point for further discussion.  Please suggest improvements!

### Proposed Resolution

[...]

Issue #173, opened 2025-09-28
Search is too Broad


mnot:
### Confirmation

- [x] I have reviewed our [existing issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

_From @chrisn in [email](https://mailarchive.ietf.org/arch/msg/ai-control/2ZbEx73js_cTQT8mI1Wy6_0ZC_s):_

While I appreciate the effort that has gone into crafting this definition, as it stands it still allows all the generative AI usage that publishers find problematic. If adopted, publishers will be in the position of having to allow the Search Category, because they cannot realistically opt-out of traditional (i.e., non generative AI search).

Given that major search engines are rapidly integrating generative AI features, a generative AI response to a user query that includes links provided via RAG or grounding could be interpreted as "directs users to the location from which the assets were retrieved". The qualifying statements: "Only those parts of applications that direct users to the location of an asset are included in this category of use. This includes the use of titles or excerpts from assets that are used to help users select between multiple candidate options" don't address this concern. The same applies to "Parts of applications that do not direct users to the location of assets, such as summaries, are not covered by this category of use", because search engines that integrate generative AI do produce summaries.

mnot:
From Bradley Silver in [email](https://mailarchive.ietf.org/arch/msg/ai-control/zaXD4Ah5ruMVIr1sbdwMVeYkOIk):

On search, I agree with the concerns Chris raised - but I also wonder whether this is something that can be better clarified in the drafting, versus there being actual disagreement about what that category is intended to cover or not, when it comes to the line between search results and "answers" with citations.

chrisn:
Here's a proposed change to the definition of Search category, which focuses its scope a bit:

The use of one or more assets to create a search index for the purpose of providing search results in the form of returning links to a website's contents. Use includes:
* Backend Processing: The internal processing of assets to perform indexing, ranking, and retrieval may utilize AI/ML models specifically and only to the extent necessary for those search functions.
* Display and Output: The presentation of the asset in the search output must include a direct reference or link to the original location. Verbatim excerpts or snippets may be displayed to assist users in evaluating the relevance of the result.

~~The Search category does not include uses that generate outputs which materially reduce the need to access the original asset.~~

chrisn:
Updated proposal, from Leah:

[...]

PR #199, opened 2026-04-14
(Non-generative) Search proposal synthesis

martinthomson:
This takes text from @erinsimon and @chrisn in #196 and #173 and smooshes it together somewhat awkwardly in an attempt to try to find a starting point.

There are clearly aspects of this that are duplicative, redundant, and (importantly) contested.  The general sense though is that this is forward movement and we would open issues to track those points of disagreement.


martinthomson:
After a lot of discussion at the interim, the (marginal) preference seemed to be in favor of #201 for this item.

PR #201, opened 2026-04-14
An iteration of Leah's wording of search

martinthomson:
As proposed by @chrisn in https://github.com/ietf-wg-aipref/drafts/issues/173#issuecomment-4246716167

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -268,7 +268,7 @@ This category of use only applies under the following conditions:
   includes a direct reference or link
   to the original location from which the asset was retrieved.
 
-* Only \[verbatim\] excerpts from the source asset are displayed
+* Only excerpts from the source asset are displayed
   to assist users in evaluating the relevance of the result.
 
 This category does not include the use of assets
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -254,10 +254,9 @@ for a narrower set of use cases.
 
 ## Search {#search}
 
-Use of an asset in applications
-where the primary purpose of that application
-is to select assets
-and direct users to the location of those assets.
+Use of an asset in an application
+that selects assets
+and directs users to the location of those assets.
 
 This category of use only applies under the following conditions:
 
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -268,8 +268,8 @@ This category of use only applies under the following conditions:
   includes a direct reference or link
   to the original location from which the asset was retrieved.
 
-* Excerpts from the asset are displayed
-  to assist users
+* When excerpts from the asset are displayed
+  they serve to assist users
   in evaluating the relevance of the result.
 
 This category does not include the use of assets
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -274,15 +274,15 @@ This category of use only applies under the following conditions:
 This category does not include the use of assets
 to generate summaries.
 
-\[Non-substantive transformations to the presentation
+\[Non-substantive changes to the presentation
 of titles or excerpts from the original assets
 is included for the purposes of accessibility.
 Translation, transcription, or text-to-speech
-are examples of non-substantive transformations
+are examples of non-substantive changes
 that could help users understand what is being presented.
 Where existing controls restrict presentation of these items,
 such as limitations on snippet size,
-those apply before any transformation.\]
+those apply before any changes.\]
 
 A preference to allow this category of use
 also implies a preference to allow any internal processing
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -118,10 +118,6 @@ Machine Learning (ML):
   to produce or improve a model that encodes the relationship
   between the data and human-defined objectives.
 
-Search Application:
-: A search application is a system that enables users
-  locate items on the internet or in a specific data store.
-
 
 # Statements of Preference {#model}
 
@@ -258,8 +254,10 @@ for a narrower set of use cases.
 
 ## Search {#search}
 
-The use of one or more assets in a search application
-that direct users to the location of those assets.
+Use of an asset in applications
+where the primary purpose of that application
+is to select assets
+and direct users to the location of those assets.
 
 This category of use only applies under the following conditions:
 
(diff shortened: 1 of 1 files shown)


tymrtn:
Nice iteration on Leah's text. 

A few suggested changes, with rationale, followed by a fully incorporated version:

Suggested changes:

- Broaden the opening definition to plural: "to direct users to the locations of those assets". The singular "location of the asset" could be read to exclude multi-result search pages where users survey several options rather than being directed to one.

- Tighten model exclusivity constraint: "exclusively by the search application and exclusively for purposes of search as defined in this category". The current "used exclusively by the search application" is necessary but not sufficient; a model could be used by the search app for non-search purposes. Double-locking to both the application and the purpose closes that gap.

- Remove the "evaluating an asset" sentence: "That is, evaluating an asset for potential use is included" is orphaned from its antecedent and could open a loophole where any amount of processing is characterized as "evaluation." Better tracked as a separate issue.

- Broaden the summary exclusion back toward Leah's original: "summaries or other new content beyond verbatim excerpts or snippets". Naming only "summaries" invites recharacterization ("it's a synthesis, not a summary"). The broader phrasing from Leah's proposal in #173 is more defensive.

- Restore "verbatim": The old text required excerpts "drawn verbatim," which this iteration dropped. Without it, "snippets", a term commonly used for AI-generated preview text, could be read to permit generative output, undermining the core goal.

- Add a substitution test: Chris's original proposal in #173 included the principle that search should not "generate outputs which materially reduce the need to access the original asset." Framing this as a substitution test aligns with established copyright concepts (market substitution under the fourth fair use factor) and captures the underlying policy goal more directly than any specific list of prohibited output types.

Fully incorporated version:

```

## Search {#search}

The use of one or more assets in a search application to direct users to the locations of those assets.

This category of use has the following conditions:

* The presentation of an asset in search output  must include a direct reference or link to the original location from which the asset was retrieved.

* To assist users in evaluating the relevance of a result,  verbatim excerpts or snippets may be displayed  from the source asset.

The internal processing of assets to perform tasks such as indexing, ranking, and retrieval may involve the use and training of AI/ML models, provided that those models are used exclusively by the search application and exclusively for purposes of search as defined in this category. 

This category does not include the use of assets to generate summaries or other new content beyond verbatim excerpts or snippets as described above. More generally, this category does not include uses that generate outputs which serve as a substitute for the original asset.

```

martinthomson:
Hi @tymrtn, this is a helpful contribution.  A few things:

* The "verbatim" question is contested and so that's not something we can resolve (see #202 as a placeholder for that discussion).
* In discussion in the meeting, we agreed to remove the "exclusion" on the summaries piece.  The rationale for that was that it was duplicative and some people felt that repeating it created the impression that it would confuse things.
* The substitution text was highly contested and Chris was very quick to suggest we split that out into a separate discussion rather than include it in this text.

I've attempted to take your suggestions, then I've iterated further.  I realized that the two bullet points were not parallel.  There is a restriction (you have to include a link) and text that expands what is permitted beyond that.  Presenting that as parallel restrictions isn't really helping comprehension.

To that end, I have a simpler presentation in mind.  Take a look at what I've pushed and see if that is clearer.  Obviously, we still need to resolve the questions about a) verbatim excerpts and b) substitutive use.

draft-ietf-aipref-vocab.md
```diff
@@ -258,32 +258,27 @@ for a narrower set of use cases.
...
+includes a direct reference or link
+to the original location from which the asset was retrieved.
+
+To assist users in evaluating the relevance of the result,
+\[verbatim\] excerpts or snippets from the source asset
+can be displayed.
+This category does not include the use of assets
+to generate summaries.
+
+This preference does not apply to any internal processing
```
MikeBishop: "does not apply to" can be read either way. Perhaps "also includes"?

meredithjacob:
From Erin's comment this version is missing "Excerpts can additionally be subject to non-substantive modifications to the presentation of material, which includes transcription, translation, text-to-speech, and speech-to-text."

If we use this as a starting point, it should include that sentence - even if there's not agreement on this now, so it's not lost for the discussion going forward.

sureshkrishnan:
At the Toronto interim there was a slight preference to use this proposal (in comparison to #199) as the starting point for the search definition. This will be followed up on list.

l-romm:
@martinthomson what if we restructure like this?

This category of use only applies where the presentation of an asset in search output includes a direct reference or link to the original location from which the asset was retrieved.

The following are permissible implementations of this use:
- Only [TBD verbatim] excerpts or snippets from the source asset to assist users in evaluating the relevance of the result are displayed.
- Non-substantive changes to the presentation of titles, excerpts, or snippets from assets, including transformations like transcription, translation, or text-to-speech that enable users to understand the material.

A preference to allow this category of use implies permitting internal processing, including the use and training of AI models, that is performed on assets by the search application, provided that the asset and the outputs of any processing is exclusively used by the search application according to the other restrictions in this section.

The following are prohibited implementations of this use:
- The use of assets to generate summaries [TBD LANGUAGE].

draft-ietf-aipref-vocab.md
chrisn: I think the intent here was "doesn't include"

draft-ietf-aipref-vocab.md
```diff
@@ -258,32 +258,41 @@ for a narrower set of use cases.
...
+* Where the presentation of an asset in search output --
+  if selected for presentation --
+  includes a direct reference or link
+  to the original location from which the asset was retrieved.
+
+* Only excerpts from the asset are displayed
+  to assist users in evaluating the relevance of the result.
+
+This category does not include the use of assets
+to generate summaries.
```
llemeurfr, line 274: this is where more details about what summaries are would be useful: "... to generate summaries of the search results as a whole, or of each individual search result".

draft-ietf-aipref-vocab.md
```diff
@@ -258,32 +258,43 @@ for a narrower set of use cases.
...
+that direct users to the location of those assets.
+
+This category of use only applies under the following conditions:
+
+* Where the presentation of an asset in search output --
+  if selected for presentation --
+  includes a direct reference or link
+  to the original location from which the asset was retrieved.
+
+* When excerpts from the asset are displayed
```
lrosenthol, line 269:
```suggestion
* When excerpts from the asset are displayed,
```

draft-ietf-aipref-vocab.md
```diff
@@ -258,32 +258,43 @@ for a narrower set of use cases.
...
+  they are only used to assist users
```
```suggestion
  they serve to assist users
```

Issue #160, opened 2025-09-13
3.2 Respecting Preferences


mnot:
### Confirmation

- [x] I have reviewed our [existing WLGC issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen%20label%3Awglc) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

Per thread on the list, there is still contention about the text in 3.2.

### Proposed Resolution

_Sourced from e-mail by [Alissa](https://mailarchive.ietf.org/arch/msg/ai-control/5vUslBkhFk-na7geaO_XFaWqc0c)_

1. Getting rid of the normative MAY. I know this was discussed on the PR thread before the -03 was published. This whole section is in the realm of non-technical considerations and using a normative keyword muddies that.

2.  Significantly paring back the text starting from “Usage preferences can be ignored …” so that it provides less fodder for future debate about what was really intended here. I have a suggestion below that echoes back to Lila’s original note about historic, widespread use of certain tools. I nixed “free expression” from the list as it felt like the everything category, but I think including the other examples is helpfully illustrative.

> Parties might choose not to honor preferences they receive on the basis of express agreements they have with other parties, for legal compliance purposes, or for other reasons. Historically, entities engaged in automated processing of digital assets in support of safety, education, scholarship, research, preservation, interoperability, accessibility, and other priorities have needed to make such choices in response to existing preference signaling mechanisms, and affected parties have had to grapple with the consequences. The preference statements described in this specification are no different.

mcerdancch-png:
I would like to highlight the importance of explicitly including cognitive accessibility preferences within this section.

People with cognitive and learning disabilities often need preferences such as:
– simplified or plain language,
– anticipation of steps in a process,
– additional response time,
– support with visual aids (icons, pictograms, real images).

These preferences must be respected in the same way as security or privacy preferences, because they directly affect users’ ability to interact with AI systems.

As an example, my research with public administrations in Barcelona resulted in a set of practical communication guidelines (Decàleg per a l’atenció al públic amb TEA, Cerdan-Chiscano, 2023, Diputació de Barcelona & Aprenem Autisme, available at: [ResearchGate link](https://www.researchgate.net/publication/395700660_Decaleg_pautes_atencio_al_public_amb_TEA)
). These show how respecting such preferences significantly improves usability for autistic people and others with cognitive disabilities.

[...]

PR #203, opened 2026-04-15
Remove the "MAY" from S3.2

martinthomson:
This is a partial response to Alissa's comments in #160.

This is not a normative statement anyway, so avoid the use of normative language.


martinthomson:
Closing in favor of #206.

PR #205, opened 2026-04-15
Switch from "ignore" to focus on a choice not to honor preferences

### LINKED ISSUE #160: 3.2 Respecting Preferences

martinthomson:
This is a lighter-weight approach to addressing the comments raised in #160.  It retains the longer list of reasons that people might choose not to honor preferences.

Closes #160.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -194,9 +194,10 @@ legal, contractual, or otherwise—might enforce stated preferences
 and thereby determine the consequences of following or not following
 a stated preference.
 
-An entity that receives usage preferences MAY choose to respect
-those preferences it has discovered, according to
-an understanding of how the asset is used,
+An entity that receives usage preferences has a choice
+about whether to respect the preferences it has discovered.
+It makes this choice according to its understanding
+of how the asset is used,
 how that usage corresponds to the usage categories
 where preferences have been stated,
 and the applicable legal context.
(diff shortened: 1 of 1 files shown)


martinthomson:
Closing in favor of the text discussed at the interim in #206.

PR #206, opened 2026-04-15
Synthesis of proposed text for issue 160

### LINKED ISSUE #160: 3.2 Respecting Preferences

martinthomson:
This is an attempt to deal with the questions around the text in
Section 3.2.  The changes here are stacked up, but really the best way
to look at this as a complete replacement for the section.

I have taken text from both Alissa and Brad in putting this together.  I
*think* that I have captured a lot of the intent behind each, without
losing things from the original text that was adding value and did not
receive objections.  This also takes Kevin's suggestion that we avoid
purely "positive" examples of use; to that end, I've added some
"negative" examples to the list.

Closes #160.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -194,9 +194,10 @@ legal, contractual, or otherwise—might enforce stated preferences
 and thereby determine the consequences of following or not following
 a stated preference.
 
-An entity that receives usage preferences MAY choose to respect
-those preferences it has discovered, according to
-an understanding of how the asset is used,
+An entity that receives usage preferences has a choice
+about whether to respect the preferences it has discovered.
+It makes this choice according to its understanding
+of how the asset is used,
 how that usage corresponds to the usage categories
 where preferences have been stated,
 and the applicable legal context.
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -193,8 +193,7 @@ This specification only seeks to ensure that preferences can be understood,
 not provide a means of ensuring that preferences are respected.
 
 Enforcement is not provided by this specification.
-Preferences do not themselves create rights, obligations, or prohibitions,
-either in the positive or the negative.
+Preferences do not themselves create rights, obligations, or prohibitions.
 Other mechanisms --
 technical, legal, contractual, or otherwise --
 might enforce adherence or non-adherance to these preferences
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -191,6 +191,8 @@ or their architecture.
 
 This specification only seeks to ensure that preferences can be understood,
 not provide a means of ensuring that preferences are respected.
+There may be considerations that take precedence over any stated preferences,
+especially when taking the public interest into account.
 
 Enforcement is not provided by this specification.
 Preferences do not themselves create rights, obligations, or prohibitions.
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -184,56 +184,42 @@ This specification provides a set of definitions for different
...
-either in the positive or the negative. Other mechanisms—technical,
-legal, contractual, or otherwise—might enforce stated preferences
-and thereby determine the consequences of following or not following
+The categories of use in the vocabulary (see {{vocab}})
+describe concrete, observable outcomes that depend on the use of assets,
+they seek to avoid describing internal details of implementations
+or their architecture.
+
+This specification only seeks to ensure that preferences can be understood,
+not provide a means of ensuring that preferences are followed.
```
```suggestion
not provide a means of ensuring that preferences are respected.
```

followed seems like the wrong word here...

draft-ietf-aipref-vocab.md
```diff
@@ -184,56 +184,42 @@ This specification provides a set of definitions for different
...
+
+Enforcement is not provided by this specification.
+Preferences do not themselves create rights, obligations, or prohibitions,
+either in the positive or the negative.
+Other mechanisms --
+technical, legal, contractual, or otherwise --
+might enforce adherence to these preferences
+and thereby determine the consequences of not following
```
```suggestion
and thereby determine the consequences of a system not respecting
```

draft-ietf-aipref-vocab.md
```diff
@@ -184,56 +184,42 @@ This specification provides a set of definitions for different
...
 a stated preference.
 
-An entity that receives usage preferences MAY choose to respect
-those preferences it has discovered, according to
-an understanding of how the asset is used,
+An entity that receives usage preferences has a choice
+about whether to respect the preferences it has discovered.
+It makes this choice according to its understanding
+of how the asset is used,
```
lrosenthol, line 208:
```suggestion
of how the asset is being used,
```

draft-ietf-aipref-vocab.md
```diff
@@ -184,56 +184,42 @@ This specification provides a set of definitions for different
...
-
-Because enforcement is not provided by this specification,
-the consequences of ignoring preferences could vary
-depending upon how a given legal jurisdiction recognizes preferences.
+Historically, entities have used assets for a variety of reasons.
+This has included a wide array of uses in applications
+including safety, advertising, education, scholarship, research,
+market intelligence, preservation, surveillance, interoperability, and accessibility.
+Those decisions were based on a subjective analysis
+of preexisting signals about preferences
```
```suggestion
of pre-existing signals about preferences
```

draft-ietf-aipref-vocab.md
```diff
@@ -184,56 +184,42 @@ This specification provides a set of definitions for different
...
+and an understanding of how the asset would be used in that application.
+The mechanisms in this document might be more direct
+than previous ways of expressing preferences,
+but the same fundamental challenges remain
+when deciding whether to use an asset.
```
```suggestion
when a system decides whether to use an asset.
```

lrosenthol (changes requested):
editorials and flagging a point about to whom this is directed...

draft-ietf-aipref-vocab.md
martinthomson: I don't think this is an improvement.

draft-ietf-aipref-vocab.md
martinthomson: My dictionary said that this was OK without the hyphen, but I've made the change.

draft-ietf-aipref-vocab.md
martinthomson: I'm OK with the passive voice here.

PR #207, opened 2026-04-15
Replace foundation model production with AI training

### LINKED ISSUE #198: Foundation Models are too specific
In Toronto, Paul expresses concern that FM excludes use cases that people want to express preferences about, and suggests broadening so that there isn't a distinction.

martinthomson:
(This commit is just the main text, I now need to go through and clean up other mentions...)

Closes #198.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -241,19 +241,12 @@ depending upon how a given legal jurisdiction recognizes preferences.
 This section defines the categories of use in the vocabulary.
 
 
-## Foundation Model Production Category {#train-ai}
+## AI Model Training Category {#train-ai}
 
-The act of using an asset to train or fine-tune a foundation model.
-
-Foundation models are large models that are produced using deep learning
-or other machine learning techniques.
-Foundation models are trained on very large numbers of assets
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -241,19 +241,12 @@ depending upon how a given legal jurisdiction recognizes preferences.
...
-Foundation models are trained on very large numbers of assets
-so that they can be applied to a wide range of use cases.
-Foundation models typically possess generative capabilities
-in one or more media.
-
-Fine-tuning can specialize a general-purpose foundation model
-for a narrower set of use cases.
+The act of using an asset
+in the production or refinement of the weights of an AI model
+that can generate content in one or more modalities
```
lrosenthol, line 248:
```suggestion
that can generate assets in one or more modalities
```

(or whatever word we land on)

martinthomson: The choice to use a different word is deliberate, because we don't want to restrict things to just things that produce entire "assets", which is a word we are using to refer to an entire artifact/thing/etc.  Here, we are looking at something more general.

paul2keller (approved):
This looks ready to merge to me

PR #209, opened 2026-04-20
Flag consensus status

mnot:


draft-ietf-aipref-vocab.md
```diff
@@ -227,6 +232,8 @@ and the applicable legal context.
 
 This section defines the categories of use in the vocabulary.
 
+NOTE: This section does not yet have consensus; see "Note to Readers" above.
```
paul2keller, line 233: I would suggest moving this up above line 228/233 which makes it more clear what the "this section" applies to.

paul2keller (changes requested):

paul2keller (approved):
LGTM

martinthomson (approved):
Much less damage than anticipated. Are you sure that you don’t want more of these notes?

mnot:
If you want to add more, please go ahead...

martinthomson:
We can start here. As we drive down into specific terms, we can split out diversify.

Issue #172, opened 2025-09-28
Missing Category for RAG and Grounding


mnot:
### Confirmation

- [x] I have reviewed our [existing issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

_From @chrisn in [email](https://mailarchive.ietf.org/arch/msg/ai-control/2ZbEx73js_cTQT8mI1Wy6_0ZC_s):_

Without a category that covers RAG and grounding usage, the vocabulary does not meet publisher needs. This need was raised at the outset of this work, in the IAB workshop in Washington DC in 2024, and included in the WG charter scope.

In addition to reducing web traffic to original content, such usage introduces the possibility of creating misleading or biased statements, an issue that impacts news publishers in particular, risking undermining their credibility. Research published by the BBC goes into this in more detail, demonstrating that AI-generated answers summarise and present content in a way that can introduce factual inaccuracies and biases that are not present in the original published content.

chrisn:
Thanks, not yet, so I need to review this in the context of the latest draft.

paul2keller:
here is a proposal for a category that I have been working on based on some side conversations. This builds on the way `ai-input` is defined in the [RSL](https://rslstandard.org/rsl#_3-4-1-1-usage-vocabulary) and [Cloudflare](https://contentsignals.org) vocabularies (which cover RAG), but added an exclusion for assets that are directly provided by the end user of the system as a user rights safeguard:

> `AI-input` Retrieving assets as input into an AI-based system to generate outputs that are presented to clients of that system (retrieval augmented generation). This does not include assets that are provided as input by end users of the AI-based system either directly or by providing the location of the asset(s).

lrosenthol:
> This does not include assets that are provided as input by end users of the AI-based system either directly or by providing the location of the asset(s).

I have significant objections to this sentence.  Adobe's Creative customers have a strong desire to be able to have their asset's associated preferences work in applications when directly presented as well.  For that reason, Adobe's applications (and services) today actively respect the `cawg.ai_inference ` preference when present in a CAWG training & data mining assertion (https://cawg.io/training-and-data-mining/1.1/).  That functionality is important to our customers and needs to be permitted.

cantechit:
This category doesn't address RAG for the purposes of training use.    If I could suggest text similar to.. 

[...]

PR #210, opened 2026-05-14
vocab: add rag-ground (RAG and grounding) category

new-world-coder:
Refs #172

This PR adds a single vocabulary category `rag-ground` ("Retrieval-Augmented Generation and Grounding") to `draft-ietf-aipref-vocab.md`, as a minimal concrete proposal for the discussion in #172.

**Design choices (minimal / single category):**
- Definition is **retrieval-mechanism-neutral** (on-demand fetch, pre-built index/corpus, or mixed), consistent with the latest clarification on #172 between automatic retrieval and grounding vs wording that implied "fetch from the web" only.
- **Distinct from** `train-ai` (inference-time retrieval/grounding only).
- **Narrower than / not implied by** `search`: allowing `search` does not imply allowing `rag-ground`; combining rules in {{combine}} still apply.

**Files:** vocabulary section + labels table (`{#t-category-labels}`).

Feedback welcome on naming (`rag-ground` vs alternatives discussed in the issue) and boundaries vs `search` / future inference categories.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -289,6 +289,33 @@ are used exclusively
 in ways that meet the above conditions.
 
 
+## Retrieval-Augmented Generation and Grounding {#rag-ground}
+
+The act of an automated system
+retrieving information from an asset,
+selecting information from a collection of assets or derived indexes,
+or otherwise incorporating asset content
+into the input of a trained AI model,
+in order to augment or ground the model's output,
+as part of operating the system
+(for example when responding to a user prompt).
+
+This category applies regardless of whether retrieval is performed on demand,
+from a pre-built index or corpus,
+or from a combination of sources.
+
+This category does not include
+the use of an asset in the production or refinement of an AI model;
+see {{train-ai}}.
+Uses that are solely for {{search}} as defined in this document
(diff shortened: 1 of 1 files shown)


new-world-coder:
Summary for reviewers:

- One new section `## Retrieval-Augmented Generation and Grounding {#rag-ground}` inserted after `## Search {#search}`.
- Labels table updated with `rag-ground`.
- Intent: satisfy **#172**'s ask for a **concrete** proposal while staying minimal (single category) so the WG can iterate or split later if consensus favors multiple inference-related labels.

Explicitly **Refs #172** (does not claim to close the issue until editors/chairs agree).

llemeurfr:
The term "rag" was discussed previously by the group members and is considered a current technique that may be replaced by other techniques in the future. The vocabulary we try to create should be independent of technical terms.

paul2keller (commented):
I think this can be closed without merging, seems like this discussion has moved elsewhere (#216)

martinthomson:
Based on discussions at our last interim, I'm going to close this.  We're focusing our effort on a fresh set of definitions now that draw on the experience here, but we want to keep the number of open threads to track to just a few.

PR #211, opened 2026-06-23
Refine section 4.1 based on Toronto discussion

kelleyk:
Here are the proposed tweaks to the definition of model training that I mentioned on the interim call.  (This was originally homework from Toronto; sorry that it's taken so long to post.)  I think that this gets us to a description of what it means to train a general-purpose foundation model without relying on any of the words that have raised objections in past drafts.

The most substantive change I'm proposing in this PR is that we move from talking about a model that "can generate" to one that "is used, or made available for use, to generate."  One of the things that I heard most clearly in Toronto from people thinking about declaring preferences was concern about the subjectivity of definitions based on purpose and intent, and about the possibility that either might change over time.

This makes the category slightly more narrow, but also more objective—something either was used or it wasn't; much less is subject to interpretation.  The same property should let implementers apply the category consistently, too.

This approach also focuses on the activity rather than the technology, which I hope addresses the concerns that were raised in Toronto about unintended consequences.

I've expanded "used" to "used, or made available for use" in an attempt to address the concerns that were raised about open-weight models and organizational boundaries.

I've replaced the "produce or refine" language in the current draft with "modify", which I think encompasses both; producing a model is the process of iteratively refining its weights.

I've also kept "asset" from the current draft.  I know that some folks had concerns about that term in Toronto; I don't feel strongly about it, so long as we have a word for "piece of data".

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -237,7 +237,7 @@ This section defines the categories of use in the vocabulary.
 
 ## AI Model Training {#train-ai}
 
-Using an asset to modify the common parameters of an AI model
+Using an asset to modify the learned parameters of an AI model
 that is used
 to generate synthetic content in one or more modalities.
 
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -237,11 +237,9 @@ This section defines the categories of use in the vocabulary.
 
 ## AI Model Training {#train-ai}
 
-The act of using an asset
-in the production or refinement of an AI model
-that can generate content in one or more modalities
-(text, image, audio, etc...).
-
+Using an asset to modify the common parameters of an AI model
+that is used, or made available for use,
+to generate synthetic content in one or more modalities.
 
 ## Search {#search}
 
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -238,7 +238,7 @@ This section defines the categories of use in the vocabulary.
 ## AI Model Training {#train-ai}
 
 Using an asset to modify the common parameters of an AI model
-that is used, or made available for use,
+that is used
 to generate synthetic content in one or more modalities.
 
 ## Search {#search}


draft-ietf-aipref-vocab.md
```diff
@@ -237,11 +237,9 @@ This section defines the categories of use in the vocabulary.
 
 ## AI Model Training {#train-ai}
 
-The act of using an asset
-in the production or refinement of an AI model
-that can generate content in one or more modalities
-(text, image, audio, etc...).
-
+Using an asset to modify the common parameters of an AI model
```
lrosenthol: I thought I would ask my AI what this meant to it - and it said:
>"Common parameters of an AI model" refers to the configurable settings that control how an AI model behaves when generating a response. These are not the internal learned weights of the model itself, but rather the runtime knobs you adjust when making a request.

That is not what we mean by training.  We mean the actual "internal learned weights".

draft-ietf-aipref-vocab.md
```diff
@@ -237,11 +237,9 @@ This section defines the categories of use in the vocabulary.
...
+that is used, or made available for use,
+to generate synthetic content in one or more modalities.
```
lrosenthol, line 242:
```suggestion
to generate content in one or more modalities.
```

synthetic adds a word that we don't use elsewhere and doesn't offer any advantages in this context

lrosenthol (dismissed):

draft-ietf-aipref-vocab.md
B1ueD0g: I agree with this concern. "Common parameters" is easy to read as runtime configuration (temperature, sampling, system-level knobs), not the learned parameters that training changes. If the intent is the training/fine-tuning boundary, something like "internal learned parameters" or "persistent model parameters" would make the distinction clearer and avoid pulling request-time configuration, retrieval state, or prompt/context construction into this category.

OR13:
```suggestion
Using an asset to modify an AI model
```

B1ueD0g:
A small framing point from the list discussion: I think it helps to keep two questions separate here:

1. the vocabulary category: whether the asset is used to modify model state that can later be used to generate content; and
2. the propagation/attachment question: whether preference information has to travel with a dataset or model artifact so a downstream actor can evaluate it.

For this PR, I would keep section 4.1 mostly on the first question. The open-weight case is important, but if "made available for use" carries too much weight, it can start to sound like model publication itself triggers the preference even where the downstream actor does not have enough attached context to evaluate it.

One possible direction would be something like:

> Using an asset to modify an AI model that is used, or made available for use by another actor, to generate content in one or more modalities.

Then, if useful, a short note could say that passing preference information with datasets or model artifacts is a processing/metadata question, not something this vocabulary definition by itself specifies.

That keeps the definition objective, while not forcing this PR to resolve the larger chain-of-custody question.

lrosenthol:
> 2. the propagation/attachment question: whether preference information has to travel with a dataset or model artifact so a downstream actor can evaluate it.
> 
That's a completely separate document.

This document is ONLY the Vocab.

B1ueD0g:
Agreed; keeping this PR strictly on the vocabulary seems right.

With that boundary, I would narrow the proposed definition to the model-state change itself and avoid importing attachment/propagation semantics here. One possible text shape would be:

> Using an asset to modify an AI model's persistent internal parameters, where the model is used, or made available for use, to generate content in one or more modalities.

If "made available for use" is still considered to pull in the downstream-actor problem, I would prefer dropping that clause here and leaving the A-to-B allocation entirely to the separate attachment/processing document. The key vocabulary distinction, for me, is persistent internal model-state modification, not request-time configuration, retrieval context, or later inference handling.

kelleyk:
I've updated this PR based on the discussion on the list.  Folks there generally agreed with the idea behind the "or made available for use" phrasing, but felt that it would be better handled as an attachment concern.  That seems just as good to me.

draft-ietf-aipref-vocab.md
martinthomson: Based on discussion, could this be something like this?

```suggestion
Using an asset to modify the learned parameters of an AI model
```

martinthomson (approved):
🤌🤌🤌

thunderpoot (approved):
🤌🤌🤌🤌

draft-ietf-aipref-vocab.md
bumblefudge: maybe adding a definition of "synthetic" to the list of terms at the end of section 2 would help, if it's confusing to only use the term here?

thunderpoot: As MT said in the plenary chat:
> > synthetic /sĭn-thĕt′ĭk/ adjective
> > Relating to, involving, or of the nature of synthesis.
> > Produced by synthesis, especially not of natural origin.
> > Prepared or made artificially.

> Seems like we can rely on the dictionary definition for the word.

chrisn: I think we should remove the word "synthetic". It only raises the question of what is considered synthetic and what is not, and this question is immaterial from the perspective of websites or creators that want to express this preference.

PR #212, opened 2026-06-24
Simplify Section 3.2

chrisn:
This removes two paragraphs from Section 3.2, following discussion in the subgroup that's formed to focus on this part of the draft.

- The first reiterates what's stated elsewhere in the draft
- The second describes the categories, so belongs in Section 4

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -225,8 +225,8 @@ NOTE: This section does not yet have consensus; see "Note to Readers" above.
 
 This section defines the categories of use in the vocabulary.
 
-They describe concrete, observable outcomes that depend on the use of assets,
-they seek to avoid describing internal details of implementations
+These categories describe concrete, observable outcomes that depend on the use of assets.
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -234,6 +225,10 @@ NOTE: This section does not yet have consensus; see "Note to Readers" above.
 
 This section defines the categories of use in the vocabulary.
 
+They describe concrete, observable outcomes that depend on the use of assets,
+they seek to avoid describing internal details of implementations
```
```suggestion
These categories describe concrete, observable outcomes that depend on the use of assets.
The definitions seek to avoid describing internal details of implementations
```

This is probably my fault, but I can't read it without wanting to fix this now.

martinthomson (approved):
This is strictly editorial and I'm OK with it.  Thanks Chris.

I don't think that we lose anything: the text that is cut is elsewhere in the document.

chrisn:
Thanks, your edit looks good. This is in preparation for further changes that I'm hoping the group will propose soon.

Issue #162, opened 2025-09-22
Removing normative language: respect, ignore


Jo-Levy:
### Confirmation

- [x] I have reviewed our [existing WLGC issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen%20label%3Awglc) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

As raised in 8/25 Jlevy comment and subsequent discussions, the language and terminology embedded throughouth is normatively charged and inaccurately attributes human intentions/emotions into this vocabulary, which by its nature is directed at computer programs or algorithms.   Throughout the Vocab document, the words "respect" and "ignore" should be changed to "follow" or "not follow" preference expresssions or intructions.   

### Proposed Resolution

Solution:  In 3.2, 
-change "whether preferences are actually respected during data processing"  to "whether preferencdesare actually followed during data processing."
 - change "An entity that receives usage preferences MAY choose to respect those preferences..." to "An entity that receives usage preferences MAY choose to follow those preferences.
-change "ignored" to "not followed", "ignoring" to "not following" and "ignore" to "not follow" throughout 3.2

martinthomson:
I think that we've made enough different changes that this is no longer a concern.  #213 make the final cuts here, so that we only use "follow" and variations on that.

I just searched for the two words in question ("respect" and "ignore") and each appears twice... sort of.  The occurrences of "respect" only appear as part of "respectively", which is fine, and the occurrences of "ignore" relate to how we handle extensions (the requirement is that those are "ignored").

In short, I think we've fixed this and I'm going to close it (it's an editorial issue).

Jo-Levy:
That makes sense —I agree. 


Sent from Yahoo Mail for iPhone


On Friday, August 14, 2026, 12:42 AM, Martin Thomson ***@***.***> wrote:

martinthomson left a comment (ietf-wg-aipref/drafts#162)
I think that we've made enough different changes that this is no longer a concern. #213 make the final cuts here, so that we only use "follow" and variations on that.

I just searched for the two words in question ("respect" and "ignore") and each appears twice... sort of. The occurrences of "respect" only appear as part of "respectively", which is fine, and the occurrences of "ignore" relate to how we handle extensions (the requirement is that those are "ignored").

In short, I think we've fixed this and I'm going to close it (it's an editorial issue).

PR #213, opened 2026-07-07
Update §3.2 Applying Preferences and §4 Vocabulary Definition

TimidRobot:
**Update Sections 3.2 Applying Preferences and Section 4 Vocabulary Definition**
- Replace Sections 3.2 Applying Preferences with proposed text from subgroup, updated to incorporate Editors' suggestions.
  - Goals of the update include (non-exhaustive):
    - Simplify and ease reader comprehension
    - Provide waypoints for well-intended conscientious public actors
    - Call attention to the limitations of the Vocabulary
  - The above goals are in tension. Hopefully the proposed text strikes a good balance that minimizes misunderstanding and unintentional harms.
- Add clarifying text to Section 4 Vocabulary Definition (this PR builds on #212)
- Uses non-normative language per:
  - #162

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -225,8 +225,8 @@ NOTE: This section does not yet have consensus; see "Note to Readers" above.
 
 This section defines the categories of use in the vocabulary.
 
-They describe concrete, observable outcomes that depend on the use of assets,
-they seek to avoid describing internal details of implementations
+These categories describe concrete, observable outcomes that depend on the use of assets.
+The definitions seek to avoid describing internal details of implementations
 or their architecture.
 
 
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -197,15 +197,6 @@ or the synthesis of multiple preferences from different sources.
 
 ## Applying Preferences {#applicability}
 
-This specification provides a set of definitions for different
-categories of use, plus a system for associating simple
-preferences to each (allow, disallow, or unknown; see {{model}}).
-
-The categories of use in the vocabulary (see {{vocab}})
-describe concrete, observable outcomes that depend on the use of assets,
-they seek to avoid describing internal details of implementations
-or their architecture.
-
 This specification only seeks to ensure that preferences can be understood,
 not provide a means of ensuring that preferences are respected.
 There may be considerations that take precedence over any stated preferences,
@@ -233,6 +224,10 @@ and the applicable legal context.
 NOTE: This section does not yet have consensus; see "Note to Readers" above.
 
 This section defines the categories of use in the vocabulary.
+
+They describe concrete, observable outcomes that depend on the use of assets,
+they seek to avoid describing internal details of implementations
+or their architecture.
 
 
 ## AI Model Training {#train-ai}
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -198,7 +198,7 @@ or the synthesis of multiple preferences from different sources.
 ## Applying Preferences {#applicability}
 
 This specification enables the expression of a defined set of preferences that
-can be communicated and interoperably understood.  Readers of this
+can be communicated and interoperably understood. Readers of this
 specification should understand that it does not:
 
 - provide for enforcement;
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -209,8 +209,9 @@ specification should understand that it does not:
   expressed preferences.
 
 An entity that receives usage preferences has a choice whether to follow those
-preferences. Multiple factors could influence that decision, but the decision
-itself is outside the scope of this document.
+preferences. This specification does not determine how that choice is made.
+Whether and under which circumstances a preference is followed is outside the
+scope of this specification.
 
 
 # Vocabulary Definition {#vocab}
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -197,26 +197,20 @@ or the synthesis of multiple preferences from different sources.
 
 ## Applying Preferences {#applicability}
 
-This specification only seeks to ensure that preferences can be understood,
-not provide a means of ensuring that preferences are respected.
-There may be considerations that take precedence over any stated preferences,
-especially when taking the public interest into account.
+This specification enables the expression of a defined set of preferences that
+can be communicated and interoperably understood.  Readers of this
+specification should understand that it does not:
 
-Enforcement is not provided by this specification.
-Preferences do not themselves create rights, obligations, or prohibitions.
-Other mechanisms --
-technical, legal, contractual, or otherwise --
-might enforce adherence or non-adherance to these preferences
-and thereby determine the consequences of not respecting
-a stated preference.
+- provide for enforcement;
+- address if, how, or when preferences should be followed or ignored;
+- address external mechanisms -- technical, legal, contractual, or otherwise --
+  that might determine adherence or non-adherence to these preferences;
+- consider situations or purposes that might justify following or ignoring
+  expressed preferences.
 
-An entity that receives usage preferences has a choice
-about whether to respect the preferences it has discovered.
-It makes this choice according to its understanding
-of how the asset is used,
-how that usage corresponds to the usage categories
-where preferences have been stated,
-and the applicable legal context.
+Because of this, stakeholders need to evaluate when and how to respect
+preferences in the context of other legal, institutional, or ethical
+commitments.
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -197,35 +197,20 @@ or the synthesis of multiple preferences from different sources.
...
-where preferences have been stated,
-and the applicable legal context.
+This specification enables the expression of a defined set of preferences that
+can be communicated and interoperably understood. Readers of this
+specification should understand that it does not:
+
+- provide for enforcement;
+- address if, how, or when preferences should be followed or ignored;
+- address external mechanisms -- technical, legal, contractual, or otherwise --
+  that might determine adherence or non-adherence to these preferences;
```
martinthomson: Is "determine" the right word here?  At some level, we don't want to imply that adherence to regulation is discretionary, but this really will come down to setting up incentives rather than imperatives.

Put differently, part of the point of this section is to emphasize that respecting preferences is -- to some extent -- discretionary.  That it might not be a choice at some level is the point of this statement, but the implication from using "determine" is that there could be a straight switch.

How about:

> 
```suggestion
- address technical, legal, contractual, or other mechanisms
  that might create a stronger requirement to follow preferences;
```

TimidRobot: this makes sense to me with the addition of "or not follow":
```diff
- - address external mechanisms -- technical, legal, contractual, or otherwise --
-   that might determine adherence or non-adherence to these preferences;
+ - address technical, legal, contractual, or other mechanisms that might create
+   a stronger requirement to follow or not follow preferences;
```

paul2keller (approved):

draft-ietf-aipref-vocab.md
```diff
@@ -197,35 +197,20 @@ or the synthesis of multiple preferences from different sources.
...
+- provide for enforcement of these preferences;
+- address if, how, or when preferences should be followed or not-followed;
+- address technical, legal, contractual, or other mechanisms that might create
+  a stronger requirement to follow or not follow preferences;
+- consider situations or purposes that might justify following or not-following
+  expressed preferences.
+
+Because of this, stakeholders need to decide when and how to follow or
+not-follow preferences in the context of other legal, institutional, or
+ethical commitments.
```
TimidRobot: Suggestion by Brandon Butler on mailing list following up on feedback from Eric Rescorla:
```suggestion
not-follow preferences in the context of legal, institutional, ethical, or
other interests and commitments.
```

Eco0099 (approved):

draft-ietf-aipref-vocab.md
```diff
@@ -197,35 +197,20 @@ or the synthesis of multiple preferences from different sources.
...
+- provide for enforcement of these preferences;
+- address if, how, or when preferences should be followed or not-followed;
+- address technical, legal, contractual, or other mechanisms that might create
+  a stronger requirement to follow or not follow preferences;
+- consider situations or purposes that might justify following or not-following
+  expressed preferences.
+
+Because of this, stakeholders need to decide when and how to follow or
+not-follow preferences in the context of legal, institutional, ethical, or
+other interests and commitments.
```
martinthomson: ~~This is my iteration on the suggestion from Sebastian on list.~~  I'll add his too.

(See below)

martinthomson: Sebastian's text, which has received considerable support on the list:

```suggestion
An entity that receives usage preferences has a choice
whether to follow those preferences.
This specification does not determine how that choice is made.
Whether and under which circumstances a preference is followed
is outside the scope of this specification.
```

mirjak: External to what? Maybe s/external/different/

martinthomson: There is also the possibility that "whether to follow" is better as "whether to follow or not follow"

martinthomson: External to the interactions between the entity with this document.  In this case, any ambiguity is fine.

draft-ietf-aipref-vocab.md
```diff
@@ -197,35 +197,20 @@ or the synthesis of multiple preferences from different sources.
...
-It makes this choice according to its understanding
-of how the asset is used,
-how that usage corresponds to the usage categories
-where preferences have been stated,
-and the applicable legal context.
+This specification enables the expression of a defined set of preferences that
+can be communicated and interoperably understood. Readers of this
+specification should understand that it does not:
+
+- provide for enforcement of these preferences;
```
mirjak: I think the original language was slightly better here:
"not provide a means of ensuring that preferences are respected"

draft-ietf-aipref-vocab.md
martinthomson: On the "external" point.

```suggestion
An entity that receives usage preferences has a choice
whether to follow those preferences.
Multiple factors could influence that decision,
but the decision itself is outside the scope of this document.
```

After consideration, though I do think that centering the focus on the document would be a good thing (to Sebastian's point during discussion), this sentence puts the document at the pointy end of the paragraph (the end) rather than the other point end (the start) and that's fine.

TimidRobot: @martinthomson I'd like to send an email to the list with the uncontested portion followed by the three proposed endings. Do you have edits you'd like to make first (as implied in meeting chat)?

martinthomson: I think that I prefer the "Multiple factors" variant here.  Other than that, this is close enough.  Thanks for asking.

draft-ietf-aipref-vocab.md
martinthomson: Thanks Mirja,

I'll let @TimidRobot decide how to deal with this, but I'm assuming that this is a friendly amendment.

TimidRobot: @mirjak if I understand your suggestion:
```suggestion
- provide a means of ensuring that preferences are respected;
```
it contains normative language ("respected").

how do you feel about:
```suggestion
- provide a means of ensuring that preferences are followed;
```
?

draft-ietf-aipref-vocab.md
achimschloss: Sorry for jumping in late - how about a pure procedural description and scoping? 

```suggestion
An entity that receives usage preferences must 
determine how to apply those preferences.
This specification does not define how that 
determination is made.
```

draft-ietf-aipref-vocab.md
martinthomson: I prefer the "followed" option, or the even more terse, but perhaps better still:

```suggestion
- ensure that preferences are followed;
```

mirjak: I don't think "respected" is normative but followed or MT's last suggest is fine. I really just think enforcement is too strong/a trigger for people assuming there should be an enforcement mechanism. We could even say "applied"; that would be even more neutral (not implying that there is a general expectation that all preferences would always be followed by "well-behaved" entities).

draft-ietf-aipref-vocab.md
mirjak: I  like that last proposal; it puts it more neutral.

TimidRobot: ~~This suggestion received the strongest support on the mailing list~~

Correction. I failed to read the "[ai-control] Re: ai-control Digest, Vol 29, Issue 50" when evaluating feedback.

PR has been updated to use Sebastian's suggestion.

TimidRobot: With the the updated text, @achimschloss 's suggestion is now:
```diff
- An entity that receives usage preferences has a choice whether to follow those
- preferences. This specification does not determine how that choice is made.
- Whether and under which circumstances a preference is followed is outside the
- scope of this specification.
+ An entity that receives usage preferences must determine how to apply those
+ preferences. This specification does not define how that determination is made.
```

draft-ietf-aipref-vocab.md
```diff
@@ -197,35 +197,21 @@ or the synthesis of multiple preferences from different sources.
...
+specification should understand that it does not:
+
+- ensure that preferences are followed;
+- address if, how, or when preferences should be followed or not-followed;
+- address technical, legal, contractual, or other mechanisms that might create
+  a stronger requirement to follow or not follow preferences;
+- consider situations or purposes that might justify following or not-following
+  expressed preferences.
+
+An entity that receives usage preferences has a choice whether to follow those
```
TimidRobot, line 211: Recommended by Andrew on mailing list (with a couple of +1s):
```suggestion
An entity that receives usage preferences may have a choice whether to follow those
```

martinthomson: I'm just going to leave this here: https://www.prindleinstitute.org/2025/04/why-follow-the-law/

PR #216, opened 2026-07-30
Add "AI System Inference" and "AI User Input" categories

natehake:
This adds two proposed inference/use vocab categories that are intended to work in tandem: 
-AI System Inference 
-AI User Input

These definitions both rely on the AI training definition of PR 211 as a logical starting point. They define inference as all use of assets by AI models beyond the modification of learned parameters. 

Together, these proposed categories would enable the expression of preferences covering the full field of post-training usage. They also give declaring parties the option to express different preferences when assets are used by autonomous systems vs when assets are identified by a human end user.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -289,6 +289,22 @@ are used exclusively
 in ways that meet the above conditions.
 
 
+## AI System Inference {#ai-system-inference}
+
+Using an asset for inference by an AI model that is used to generate synthetic content in one or more modalities. 
+Inference means all use beyond the modification of an AI model’s learned parameters. 
+This category does not include the use of assets that a human end user has specifically identified 
+(by providing either the asset itself or a location from which it can be retrieved).
+
+
+## AI User Input {#ai-user-input}
+
+Using an asset that a human end user has specifically identified 
+(by providing either the asset itself or a location from which it can be retrieved)
+for inference by an AI model that is used to generate synthetic content in one or more modalities. 
+Inference means all use beyond the modification of an AI model’s learned parameters.
+
+
 ## Vocabulary Extensions {#vocab-extension}
 
 Extensions to this vocabulary need to be defined in an RFC
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -106,6 +106,9 @@ Asset:
 Declaring party:
 : The entity that expresses a preference with regards to an Asset.
 
+Inference:
+: All use of an AI system, except use that modifies the learned parameters of an AI model.
+
 
 # Statements of Preference {#model}
 
@@ -289,20 +292,15 @@ are used exclusively
 in ways that meet the above conditions.
 
 
-## AI System Inference {#ai-system-inference}
+## Inference Directed by User {#ai-inference-user}
 
-Using an asset for inference by an AI model that is used to generate synthetic content in one or more modalities. 
-Inference means all use beyond the modification of an AI model’s learned parameters. 
-This category does not include the use of assets that a human end user has specifically identified 
-(by providing either the asset itself or a location from which it can be retrieved).
+Using an asset during inference where the asset has been explicitly selected by a human user. 
+This includes assets provided directly by the user or assets whose location has been explicitly specified by the user.
 
 
-## AI User Input {#ai-user-input}
+## Inference Directed by System {#ai-inference-system}
 
-Using an asset that a human end user has specifically identified 
-(by providing either the asset itself or a location from which it can be retrieved)
-for inference by an AI model that is used to generate synthetic content in one or more modalities. 
-Inference means all use beyond the modification of an AI model’s learned parameters.
+Using an asset during inference where the asset is selected by the system itself or any other mechanism that is not a human user.
 
 
 ## Vocabulary Extensions {#vocab-extension}


sposth:
I like the distinction between system-selected and user-selected assets, but I think the definitions can be much clearer.

The key distinction is not inference itself, but who identifies the asset used during inference.

For example:

**AI System Inference**

Using an asset during inference by an AI system where the asset is identified or retrieved by the system itself. Inference means all use beyond the modification of an AI model's learned parameters.

**AI User Input**

Using an asset during inference by an AI system where the asset has been explicitly identified by a human user. This includes assets provided directly by the user or assets whose location has been explicitly specified by the user. Inference means all use beyond the modification of an AI model's learned parameters.

This makes the distinction immediately visible:

- the system selects or retrieves the asset; or
- the human explicitly selects the asset.

The rest of the definition is common to both categories.

I also think another aspect deserves discussion.

The phrase

> "by providing either the asset itself or a location from which it can be retrieved"

does more than define the category. It also implies different attachment and discovery mechanisms.

A directly provided asset can carry an attached preference signal, for example through embedded metadata, a sidecar file, or an associated registry declaration. By contrast, when a user provides a URL or domain, the preference is most likely resolved through a location-based mechanism rather than being attached directly to the asset.

It would therefore be useful to clarify whether these are simply equivalent ways of identifying an asset, or whether the specification expects different attachment and discovery mechanisms depending on whether the user provides the asset itself or only a reference to its location. If the latter is intended, making this explicit may be worthwhile, as it has implications for implementers.

One additional observation concerns the naming. The distinction is really about who selected the asset, not about the inference process itself. For that reason, names such as **System-selected Assets** and **User-selected Assets**, or **System-selected Inference** and **User-directed Inference**, may communicate the distinction more clearly than AI User Input, which could easily be interpreted as referring to prompts rather than to the source material used during inference.

coopdanger:
1. When using the word "beyond" in the phrase "beyond the modification of an AI model's learned parameters," does "beyond" mean "after" or does "beyond" mean "except"? 

2. I think the term "inference" should be defined in Section 2 rather than here, which would allow "Inference means all use beyond the modification of an AI model's learned parameters" to be removed from this section.

3. I think Sebastian's suggested re-labeling of these two categories is useful but it might be clearer if they were symmetric: **system-directed inference** and **user-directed inference**. 

4. I think the interactions between system-directed inference and the search category will require a lot of clarification. If an expressed preference says "search=y;system-inference=n," how is that supposed to be interpreted by a search application?

natehake:
To Alissa's comments: 
-"beyond" is intended to mean "all use that doesn't change the learned parameters." So I suppose it's both after and except in that sense. I'd be fine swapping out that word if folks think another is cleaner 
-yes, defining inference makes sense. Timid Robot's proposal on the list seems good to me 
-these names are ok with me, but I'd prefer having "AI inference" in the name just for user clarity 
-yes, the way search is written it's intended to operate as an exception where a "search=y, inference =n" would permit all inference that qualifies as search (but no other inference for that category)

On Sebastian's comment: I mentioned this on the mailing list, but the reason I don't think that approach works is that the two halves do not cover the whole field. One needs to define itself by reference to the other in order for us to achieve full coverage.

paul2keller (commented):
I think it would make sense to get this PR into a state where it can serve as a concrete proposal for the discussion in London, rather than having several versions of the text scattered across the list.

My suggestion for this would be to update it to reflect @TimidRobot 's 7 August redraft: "inference" defined once in Section 2, symmetric "directed by system" / "directed by user" naming, and the "or any other mechanism that is not a human user" catch so nothing falls between the two categories. 

That version also picks up @sposth 's point that the distinction turns on who identifies the asset, which I think is right. I personally have no attachment to the wording @natehake and I originally proposed and I think this redraft answers Sebastian's naming objection and @coopdanger 's Section 2 point better than what we originally proposed. 

@TimidRobot are you ok with this?
@martinthomson does this approach make sense to you? 
@natehake since the branch is yours, I assume it's easiest if you push the update?

martinthomson:
That plan sounds good.  I'm happy to make a copy of the PR or suggestions against this one, but I think that it's best coming from @natehake and @TimidRobot.  If there is something I can do to help, let me know.

natehake:
Makes sense to me. This is my first PR, but I *think* I have correctly updated the pull request to reflect @TimidRobot's updated language. If I have done something wrong or need to do something further, please let me know @martinthomson / @paul2keller 

Note one change I made from Timid Robot's language: since the training definition has now been updated to focus on "learned parameters" I tweaked the inference definition to say: "All use of an AI system, except use that modifies the learned parameters of an AI model." I am not wedded to this specific language though, and perhaps there is a more elegant way of writing it.

coopdanger:
This is improving IMO. I think it has two key problems:

1. Nowhere do we define "system" or "AI system." And I believe this is purposeful, that we have avoided defining AI system because doing so is hard. We found ways around this in the other two category definitions by using "AI model" and "search application."

2. What is supposed to happen when agents act on the human user's behalf? For example, I ask my local agent to find five photos on my hard drive, upload them to a cloud-based AI system, and make me a holiday card featuring the photos. When my agent provides the photos to the AI system in the cloud, are those photos considered user-selected or system-selected? They aren't selected by the cloud AI system, but they are selected by the local agent system.

draft-ietf-aipref-vocab.md
```diff
@@ -289,6 +292,17 @@ are used exclusively
 in ways that meet the above conditions.
```
martinthomson, line 293: Some editorial notes:

The inference categories need to address the potential for overlap with search.  Search almost certainly involves inference and we need to be clear on which preference applies if the usage is identifiably both search and inference.  My sense is that search is a narrower notion generally - even if it includes stuff outside of inference, it's only a tiny subset of possible inference uses.  If you agree, then my suggestion would be to say that a search preference applies when the usage is clearly search (and the inference preferences do not).

As a matter of document structure, I would prefer that search be listed last (sort from most general to most specific).  That follows on from the previous point.

I'm happy to make these changes during integration, if the group decides on something like this. I just wanted to flag these.

martinthomson:
Based on discussions at our last interim, I'm going to close this. We're focusing our effort on a fresh set of definitions now that draw on the experience here, but we want to keep the number of open threads to track to just a few.

PR #219, opened 2026-08-14
Remove nesting

### LINKED ISSUE #170: Hierarchical structure is problematic and introduces unnecessary complexity
### Confirmation

### Issue Description

The hierarchical structure in vocab-03 -  placing categories like Automated Processing, Search, and AI Training in parent–child relationships - introduces complexity without real benefit. In practice, preferences can be expressed directly on each activity, so the hierarchy does not add clarity; instead it forces artificial nesting. Real-world processing often overlaps across categories, and the hierarchy has already led to a patchwork of overlapping or conflicting proposals. Because publishers simply want to state “allow” or “disallow” for specific uses or use multiple preferences, the extra layer of hierarchy is unnecessary and risks confusing both implementers and publishers. A flatter, modular vocabulary would make preferences easier to understand, more consistent to apply, and better aligned with real systems. 

### Proposed Resolution

Remove the heirarchical structure and replace with a flatter, modular structure that would make the preferences easier to understand, more consistent to apply, and better aligned with real systems.

martinthomson:
The hierarchical design we started with in earlier revisions has been clearly replaced with a system of tiling.  No term nests within another and no overlap is tolerated.  This change executes on that fully.

There might be a few minor cleanups throughout as I touched those parts, but I left the contentious stuff along.  Those changes should be purely editorial.

Closes #170.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -287,7 +287,7 @@ This approach is strongly encouraged for any extension.
 
 Systems that use this vocabulary might seek to integrate
 the terms in this vocabulary
-as part of a larger data model.
+as part of a larger data model that includes other terms not defined here.
 Such usage is not subject to the RFC requirement above,
 but special care is needed
 to avoid defining overlapping categories of use.
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -274,7 +274,9 @@ Extensions to this vocabulary are defined
 in a standards-track RFC that updates this document.
 
 The definition of the extension MUST define
-how any potential overlap is resolved.
+how any potential overlap between usage categories is resolved.
+Definitions can identify which usage category applies
+for any such overlap.
 
 This document resolves overlap between categories of use
 by explicitly designating any usage
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -278,12 +278,6 @@ how any potential overlap between usage categories is resolved.
 Definitions can identify which usage category applies
 for any such overlap.
 
-This document resolves overlap between categories of use
-by explicitly designating any usage
-that might be covered by multiple terms
-as belonging to a single category.
-This approach is strongly encouraged for any extension.
-
 Systems that use this vocabulary might seek to integrate
 the terms in this vocabulary
 as part of a larger data model that includes other terms not defined here.
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -270,7 +270,7 @@ in ways that meet the above conditions.
 
 ## Vocabulary Extensions {#vocab-extension}
 
-Extensions to this vocabulary MUST be defined
+Extensions to this vocabulary are defined
 in a standards-track RFC that updates this document.
 
 Vocabulary extensions MUST consider
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -288,7 +288,7 @@ This approach is strongly encouraged for any extension.
 Systems that use this vocabulary might seek to integrate
 the terms in this vocabulary
 as part of a larger data model.
-This is not subject to the RFC requirement above,
+Such usage is not subject to the RFC requirement above,
 but special care is needed
 to avoid defining overlapping categories of use.
 {{mapping}} describes how concepts from an alternative format
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -125,20 +125,6 @@ A statement of preference can indicate preferences
 about some, all, or none of the categories from the vocabulary.
 This can mean that the preference is unknown for some usage categories.
 
-Some categories describe a proper subset of the usages of other categories.
-A preference that is stated for the more general category applies
-if the preference is unknown for the more specific category.
-
-For example, a more general category might be assigned a preference
-that allows the associated usage.
-In the absence of any statement of preference
-regarding categories that are more specific subsets of that usage category,
-usage within those categories would be also be allowed.
-An explicit preference regarding the more specific usage category
-can be used to disallow the more specific usage,
-while indicating that other usage within the more general category
-is permissible.
-
 After processing a statement of preferences
 the recipient associates each category of use
 one of three preference values: "allowed", "disallowed", or "unknown".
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -169,6 +155,12 @@ There are also some caveats that need to be considered
 as it relates to understanding what the preferences for a given asset are
 (as opposed to what actions might then follow).
 
+A recipient can only apply preferences it understands.
+Recipients might not implement this specification.
```
mnot: Hmm. This may raise eyebrows.

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +271,26 @@ in ways that meet the above conditions.
...
+how any potential overlap is resolved.
 
-Any future extensions to this vocabulary MUST NOT introduce additional categories
-that include existing categories defined in the vocabulary.
-That is, new categories of use can be defined as a subset of an existing category,
-but not a superset.
+This document resolves overlap between categories of use
+by explicitly designating any usage
+that might be covered by multiple terms
+as belonging to a single category.
```
mnot: This is a bit abstract

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +271,26 @@ in ways that meet the above conditions.
...
+This approach is strongly encouraged for any extension.
 
-Systems that use this vocabulary might define their own extensions
+Systems that use this vocabulary might seek to integrate
+the terms in this vocabulary
 as part of a larger data model.
+This is not subject to the RFC requirement above,
```
mnot:
```suggestion
Such usage is not subject to the RFC requirement above,
```

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +271,26 @@ in ways that meet the above conditions.
...
+This document resolves overlap between categories of use
+by explicitly designating any usage
+that might be covered by multiple terms
+as belonging to a single category.
+This approach is strongly encouraged for any extension.
 
-Systems that use this vocabulary might define their own extensions
+Systems that use this vocabulary might seek to integrate
+the terms in this vocabulary
 as part of a larger data model.
```
mnot:
```suggestion
as part of a larger data model that includes other terms not defined here.
```

draft-ietf-aipref-vocab.md
ekr-ietf: I don't quite understand why we need this text. Isn't that always true?

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +271,26 @@ in ways that meet the above conditions.
 
 ## Vocabulary Extensions {#vocab-extension}
 
-Extensions to this vocabulary need to be defined in an RFC
-that updates this document.
+Extensions to this vocabulary MUST be defined
+in a standards-track RFC that updates this document.
```
ekr-ietf, line 274: I understand what you are doing here, but why not just say "are defined" rather than trying to cram in 2119 keywords.

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +271,26 @@ in ways that meet the above conditions.
...
+
+Vocabulary extensions MUST consider
+how the definition of terms could overlap with other existing terms.
```
ekr-ietf: I would remove this. I mean you should do this but it's not really actionable unless you are requiring a section in the extensions document. Otherwise you are saying extension authors need to consider.

draft-ietf-aipref-vocab.md
ekr-ietf: Agreed.

paul2keller (approved):
I think this is good to merge and it will be helpful to have a version without nesting as basis for the discussion in London.

draft-ietf-aipref-vocab.md
martinthomson:
```suggestion
```

Yeah, no need.

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +270,26 @@ in ways that meet the above conditions.
...
-that updates this document.
-
-Any future extensions to this vocabulary MUST NOT introduce additional categories
-that include existing categories defined in the vocabulary.
-That is, new categories of use can be defined as a subset of an existing category,
-but not a superset.
-
-Systems that use this vocabulary might define their own extensions
-as part of a larger data model.
+Extensions to this vocabulary MUST be defined
```
```suggestion
Extensions to this vocabulary are defined
```

draft-ietf-aipref-vocab.md
martinthomson: I believe that the usual model for extensions is that an extension doesn't generally have to update the document that defined the extension point.  In this case, the agreement was to require an RFC.  Still, I'm not a fan of overuse of MUST.

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +270,26 @@ in ways that meet the above conditions.
...
-That is, new categories of use can be defined as a subset of an existing category,
-but not a superset.
-
-Systems that use this vocabulary might define their own extensions
-as part of a larger data model.
+Extensions to this vocabulary are defined
+in a standards-track RFC that updates this document.
+
+Vocabulary extensions MUST consider
+how the definition of terms could overlap with other existing terms.
```
```suggestion
```

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +270,24 @@ in ways that meet the above conditions.
...
 
-Any future extensions to this vocabulary MUST NOT introduce additional categories
-that include existing categories defined in the vocabulary.
-That is, new categories of use can be defined as a subset of an existing category,
-but not a superset.
+This document resolves overlap between categories of use
+by explicitly designating any usage
+that might be covered by multiple terms
+as belonging to a single category.
+This approach is strongly encouraged for any extension.
```
```suggestion
how any potential overlap between usage categories is resolved.
Definitions can identify which usage category applies
when such an overlap occurs.
```

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +270,24 @@ in ways that meet the above conditions.
 
 ## Vocabulary Extensions {#vocab-extension}
 
-Extensions to this vocabulary need to be defined in an RFC
-that updates this document.
+Extensions to this vocabulary are defined
+in a standards-track RFC that updates this document.
+
+The definition of the extension MUST define
+how any potential overlap is resolved.
```
```suggestion
The definition of the extension MUST define
how any potential overlap between usage categories is resolved.
Definitions can identify which usage category applies
for any such overlap.
```

draft-ietf-aipref-vocab.md
```diff
@@ -279,16 +270,26 @@ in ways that meet the above conditions.
...
+how any potential overlap between usage categories is resolved.
+Definitions can identify which usage category applies
+for any such overlap.
+
+This document resolves overlap between categories of use
+by explicitly designating any usage
+that might be covered by multiple terms
+as belonging to a single category.
+This approach is strongly encouraged for any extension.
+
```
```suggestion
```

Issue #163, opened 2025-09-22
Replacing: "digital assets" with "digital content"


Jo-Levy:
### Confirmation

- [x] I have reviewed our [existing WLGC issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen%20label%3Awglc) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

The term "digital asset" is used throughout the Vocab document.  It is confusing and misleading.  A "digital asset" generally and commonly refers to cryptocurrency. The United States Internal Revenue Service (IRS) defines it as "any digital representation of value on a cryptographically secured, distributed ledger (blockchain) or similar technology."  See https://www.irs.gov/filing/digital-assets#description.  
 It's not a term used in robots.txt or the Attach documet which uses the term "content" and "content-usage." A preferable term would be "Digital Content."

### Proposed Resolution

Replace "digital asset" or "digital assets" with "digital content." Replace "asset" with "digital content." Replace "one or more assets" with "digital content." There are 23 instances of the term asset or assets throughout the document.  By using "digital content" there is also no new for a definition of the term "Digital Asset."

lrosenthol:
On the other hand, @Jo-Levy - standards such as C2PA and JPEG Trust both use the term "asset" or "media asset" to differentiate the entire entity/file from just the "content" - in that there exists other parts (e.g. metadata) of an "asset" beyond just the content.

See https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_assets_and_content and https://ds.jpeg.org/documents/jpegtrust/wg1n100972-105-REQ-Terms_and_definitions_for_JPEG_Trust_v2_0.pdf

mnot:
March interim:

MT: agree with Jo; 'asset' means something owns something. Since we're not talking about copyright it's a misleading term colloquially. I know that people have put a lot of time into these terms, but I disagree with the outcome. 

Lila: +1 to MT. Too specific; may be american viewpoint. Content is more all-purpose.

Farz: agree with changing term.

lrosenthol:
Just for fun, I asked a few different AI systems about alternative words for asset & content...The ones that they agreed on:

- resource
- entity
- artifact
- item

chrisn:
@lrosenthol "Resource" is bad as it carries the meaning of being there to be used, which runs counter to the goal of being able to express preferences about usage.

[...]

PR #220, opened 2026-08-14
Loosen requirement on HTTP resources

### LINKED ISSUE #168: Attachment: Section 2 preservation of HTTP request header "MUST" is too strong
I think this is a couple days late for the formal WGLC, but hope it is helpful to file this here as an issue ahead of the upcoming meeting. I can also take this to the list if prefered.

### Confirmation

### Issue Description

Section 2 of the Attachment draft has a new section about retaining the `Content-Usage` header from requests, and re-using them on responses for the same content later:

> Servers MUST retain any preferences associated with a request if the content of that request is used to answer later requests. For example, the content of a PUT request that is used to answer subsequent GET requests. Note that servers that have not been updated to understand this field will not comply with this requirement.

I think having the server retain the preference metadata is a great idea, but making it a "MUST" for any server which implements usage preferences is too restrictive. I believe servers should be allowed (and encouraged) to include this metadata header even if they don't have the means to implement state persistence based on request headers. Servers might implement alternative or more ergonomic mechanisms for expressing preferences, for instance via an HTML form field, and that is fine.

### Proposed Resolution

Make it "MAY".

Overall I think this should be more of a recommendation or best practice. Given that I would defer to the editors in terms of how best to phrase things.

martinthomson:
Plus a few minor editorial things in the attachment document, tracking similar changes in the vocabulary.

Closes #168.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -215,7 +215,7 @@ attached preferences for a PUT request,
 where the content that is used to answer subsequent GET requests,
 might be copied to the response to those GET requests.
 For this to happen,
-servers or resources need to be updated
+servers or resources would need to be updated
 to understand this field
 and then decide to use the preferences.
 
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -160,8 +160,8 @@ are intended to be complementary with other mechanisms.
 ### Embedded Preferences
 
 Embedding preferences is expected to be an effective means
-of associating preferences with content,
-because it ensures that metadata is always associated with content.
+of associating preferences with assets,
+because it ensures that metadata travels with the asset.
 This document, however, does not define any specific means of embedding preferences
 in content.
 
@@ -174,8 +174,8 @@ whether it represent audio, documents, images, video,
 or other types of content.
 Furthermore,
 some content types,
-such as plain text (`text/plain`),
-offer no standardized means of embedding preferences.
+such as plain text (`text/plain` and many other purely textual formats),
+offer no standardized and interoperable means of embedding metadata.
 
 
 ### Registry-Based Preferences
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-attach.md
```diff
@@ -160,8 +160,8 @@ are intended to be complementary with other mechanisms.
 ### Embedded Preferences
 
 Embedding preferences is expected to be an effective means
-of associating preferences with content,
-because it ensures that metadata is always associated with content.
+of associating preferences with assets,
+because it ensures that metadata travels with the asset.
```
mnot, line 164: Remember #163.

ekr-ietf, line 164:
```suggestion
because it allows metadata to travel with the asset.
```

It doesn't ensure it because it can be stripped.

draft-ietf-aipref-attach.md
```diff
@@ -207,14 +207,17 @@ that applies the representation data ({{Section 8.1 of HTTP}}).
...
 if the content of that request
 is used to answer later requests.
 For example,
-the content of a PUT request that is used
-to answer subsequent GET requests.
-Note that servers that have not been updated to understand this field
-will not comply with this requirement.
+attached preferences for a PUT request,
+where the content that is used to answer subsequent GET requests,
+might be copied to the response to those GET requests.
```
ekr-ietf, line 216: This is old text, but how would I indicate to the server that I don't want it to train on this content but that I don't want it to propagate the preference?

draft-ietf-aipref-attach.md
```diff
@@ -207,14 +207,17 @@ that applies the representation data ({{Section 8.1 of HTTP}}).
...
+For this to happen,
+servers or resources need to be updated
```
ekr-ietf:
```suggestion
servers or resources would need to be updated
```

draft-ietf-aipref-attach.md
martinthomson: I don't think that's possible.  Worth an issue?

draft-ietf-aipref-attach.md
martinthomson: Yeah, it would be good to resolve that.  The goal here is to close the distance between the two documents first.  Then at least it is easier to do a search and replace if we conclude somehow.

PR #221, opened 2026-08-14
Address distinction between access and use

### LINKED ISSUE #167: Distinguishing access and use
### Confirmation

### Issue Description

I'm going to quote @coopdanger here, because her formulation of the issue is pretty good.  I will follow up with a nit on the details, but that doesn't diminish the value of the feedback:

> This raises the question of whether the vocabulary is only meant to be used to express preferences about usage and not initial access regardless of the attachment mechanism. The very first sentence of the vocab doc uses the ambiguous term “processing” while most of the rest of the doc uses “usage.” If the intent is to define a vocabulary that only expresses preferences about an asset after the asset has been obtained (by whatever means), it needs an explicit scoping limitation to that effect. And then every attachment mechanism that gets defined in the IETF or elsewhere needs to reflect that.
>
> Making this distinction explicit might be a simple resolution to some of the concerns that have been raised about the automated processing category, because it would make clear that it’s unnecessary to resolve the question of whether crawling/accessing a site is itself considered automated processing. No preferences are being expressed about access, other than insofar as sites would expect crawlers not to crawl their site if the site wishes to disallow all the uses of interest to the crawler. It would not resolve the concerns about behavior habituation and misinterpretation, though.
>
> Even with this distinction made explicit, there would still be the possibility of basically nonsensical use cases, e.g.:
>
> ```http
> User-Agent: *
> Content-Usage: /foo bots=n
> ```
> 
> Like what is the point of allowing crawling/access but then disallowing all usage? Maybe there is a use case I’m missing.
>
> Note also that it may be entirely unrealistic to say that everywhere an attachment mechanism is specified, it has to preserve this distinction between access and usage. It’s not like we can constrain mechanisms that get defined elsewhere.
> 
> I’ve never really been convinced of the need for the top-level category, but if we are going to have one we need to be clear about when it applies, and clarifying that it only applies after an asset has been obtained might help mitigate concerns about its breadth.

### Proposed Resolution

There was some effort to make this distinction in editing, but it's clear that we need to be much clearer about it throughout.

### LINKED ISSUE #193: Carrying preferences with stuff
### Confirmation

- [x] I have reviewed our [existing issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen) and this issue is not present there.

### Issue Description

@TimidRobot points out that we don't really talk about our expectations around how preferences might be carried along with the stuff (currently, "assets") that they apply to.

I think that our presumption thus far was that anyone who acquires covered stuff carries metadata for that stuff along with it so that any use can be informed by preferences.  This means carrying robots.txt or headers, for the attachments we are using.

### Proposed Resolution

This might be tricky, but I think that our expectation is that metainformation and preferences live with the things that they cover.  That is going to be hard for things like robots.txt, which might not naturally follow along with the bytes.

martinthomson:
There's a lengthy discussion of this in #167 and #193, but the change here is short enough, I think.

This addresses questions of the difference between getting something and then using it.  It also addresses the expectations about preferences with respect to uses of assets downstream of acquisition.  This is particularly relevant to the attachment mechanisms we have here (unlike embedded metadata or registries) because the preferences are external to the asset itself.

Closes #167.
Closes #193.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -100,24 +100,6 @@ these allow for the automated gathering of preferences
 in the same way that content is obtained.
 
 
-## Acquisition and Following Preferences
-
-The acquisition of an asset --
-or the receiving of a representation --
-is distinct from how that asset is subsequently used.
-
-Consequently, this document takes the view
-that usage preferences do not apply
-to the acquisition of assets.
-They only apply to how those assets are used;
-see {{Section 3.2 of VOCAB}} for a discussion
-on what it means to follow or not follow a preference.
-
-Mechanisms for limiting access to assets,
-such as access control or the `Disallow` rules in "robots.txt",
-are out of scope for this work.
-
-
 ## Statements of Preference
 
 The format of a statement of preference
@@ -157,6 +139,45 @@ User-Agent: *
 Allow: /
 Content-Usage: train-ai=n
 ~~~
+
+
+## Acquisition and Following of Preferences
+
+The acquisition of an asset --
+or the receiving of a representation --
+is distinct from how that asset is subsequently used.
+
+Consequently, this document takes the view
+that usage preferences do not apply
+to the acquisition of assets.
+They only apply to how those assets are used;
+see {{Section 3.2 of VOCAB}} for a discussion
+on what it means to follow or not follow a preference.
+
+Mechanisms for limiting access to assets,
+such as access control or the `Disallow` rules in "robots.txt",
+are out of scope for this work.
+
+
+## Retention and Distribution of Preferences
+
+The mechanisms in this document provide usage preferences
+external to the data that comprises the asset.
+
+A system might need to apply usage preferences
+in a decision some time after obtaining the asset.
+Such a system therefore needs to retain
+any applicable preferences alongside assets
+until any decision are made.
+
+Where acquisition and use occur in separate subsystems,
+the subsystem responsible for acquisition
+needs to provide some means
+of associating the usage preferences it receives
+with the assets it makes available for use.
+A subsystem that uses an asset
+cannot assume that no preferences have been expressed
+without such a mechanism.
 
 
 ## Other Mechanisms
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -100,6 +100,24 @@ these allow for the automated gathering of preferences
 in the same way that content is obtained.
 
 
+## Acquisition and Following Preferences
+
+The acquisition of an asset --
+or the receiving of a representation --
+is distinct from how that asset is subsequently used.
+
+Consequently, this document takes the view
+that usage preferences do not apply
+to the acquisition of assets.
+They only apply to how those assets are used;
+see {{Section 3.2 of VOCAB}} for a discussion
+on what it means to follow or not follow a preference.
+
+Mechanisms for limiting access to assets,
+such as access control or the `Disallow` rules in "robots.txt",
+are out of scope for this work.
+
+
 ## Statements of Preference
 
 The format of a statement of preference
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -147,9 +147,8 @@ The acquisition of an asset --
 or the receiving of a representation --
 is distinct from how that asset is subsequently used.
 
-Consequently, this document takes the view
-that usage preferences do not apply
-to the acquisition of assets.
+Usage preferences do not apply
+to how assets are acquired.
 They only apply to how those assets are used;
 see {{Section 3.2 of VOCAB}} for a discussion
 on what it means to follow or not follow a preference.


draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,45 @@ Content-Usage: train-ai=n
 ~~~
 
 
+## Acquisition and Following of Preferences
+
+The acquisition of an asset --
+or the receiving of a representation --
```
mnot, line 147: "receipt"? "fetching"?

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,45 @@ Content-Usage: train-ai=n
...
+is distinct from how that asset is subsequently used.
+
+Consequently, this document takes the view
```
mnot: documents don't take views.

mnot:
I do wonder whether it's better to do this in the vocabulary document. Yes, it's about attachment, but this document is not the only document defining attachment; only the vocabulary document is the choke point. Fundamentally, this text is about the _nature_ of the vocabulary terms -- it just happened to be expressed in terms of how it's attached.

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,45 @@ Content-Usage: train-ai=n
...
+that usage preferences do not apply
+to the acquisition of assets.
+They only apply to how those assets are used;
+see {{Section 3.2 of VOCAB}} for a discussion
+on what it means to follow or not follow a preference.
+
+Mechanisms for limiting access to assets,
+such as access control or the `Disallow` rules in "robots.txt",
+are out of scope for this work.
```
ekr-ietf, line 158: Perhaps also "and should be enforced independently of the mechanisms defined in this document."

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,45 @@ Content-Usage: train-ai=n
...
 
+## Acquisition and Following of Preferences
+
+The acquisition of an asset --
+or the receiving of a representation --
+is distinct from how that asset is subsequently used.
+
+Consequently, this document takes the view
+that usage preferences do not apply
+to the acquisition of assets.
```
martinthomson:
```suggestion
Usage preferences do not apply
to how assets are acquired.
```

martinthomson:
Let's discuss this at the upcoming meeting.  I agree that this might be worth moving to the vocabulary document.  It's fairly fundamental.

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,44 @@ Content-Usage: train-ai=n
 ~~~
 
 
+## Acquisition and Following of Preferences
+
+The acquisition of an asset --
+or the receiving of a representation --
+is distinct from how that asset is subsequently used.
+
+Usage preferences do not apply
```
lrosenthol, line 150:
```suggestion
Usage preferences do not vary, regardless of
```

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,44 @@ Content-Usage: train-ai=n
...
+to how assets are acquired.
```
lrosenthol, line 151:
```suggestion
how assets are acquired.
```

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,44 @@ Content-Usage: train-ai=n
...
+
+Mechanisms for limiting access to assets,
+such as access control or the `Disallow` rules in "robots.txt",
+are out of scope for this work.
+
+
+## Retention and Distribution of Preferences
+
+The mechanisms in this document provide usage preferences
+external to the data that comprises the asset.
```
lrosenthol, line 164: That is NOT always the case and falls into the attachment mechanism, not vocab

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,44 @@ Content-Usage: train-ai=n
...
+
+A system might need to apply usage preferences
+in a decision some time after obtaining the asset.
+Such a system therefore needs to retain
+any applicable preferences alongside assets
```
lrosenthol, line 169:
```suggestion
any applicable preferences associated with assets
```

draft-ietf-aipref-attach.md
```diff
@@ -141,6 +141,44 @@ Content-Usage: train-ai=n
...
+until any decision are made.
+
+Where acquisition and use occur in separate subsystems,
+the subsystem responsible for acquisition
+needs to provide some means
```
lrosenthol, line 174: maybe talk about ensuring that the information is maintained?

lrosenthol (changes requested):

draft-ietf-aipref-attach.md
Jo-Levy: We made a point to remove reference to enforcement in other issues/sections. I think ekr's point is that enforcement is out of scope for this work, and I agree with that,  but I'm not 100% sure that's the intent due to the words "should be". Suggest either not adding ekr's language or change it to simply add  "and enforcment" after "robots.txt"

martinthomson:
Rolled up into the large set of changes that came out of the discussion in London.

PR #222, opened 2026-08-14
Rework attachment abstract

### LINKED ISSUE #165: "consumed by automated systems"
### Confirmation

### Issue Description

 In the abstract,  the phrase “Content creators and other stakeholders might wish to signal their preferences about how their content might be consumed by automated systems” is confusing and does not accurately describe the scope. The preference is not about how  content is consumed from the website.  The preference is about how it is  used after it is accessed or consumed. Also, the phrase "automated systems" is overly broad, particularly since here it seems to refer to crawlers but later it seems to refer to AI systems that need to read the preferences after it is collected. This will lead to confusion.

Suggestion:  change to "about whether website content that is accessed orconsumed by crawlers may later be used to train or develop AI models."

### Proposed Resolution

Suggestion:  change to "about whether website content that is accessed orconsumed by crawlers may later be used to train or develop AI mode

martinthomson:
This takes a more dramatic cut than proposed in the issue, but I think it is best to focus on this document being strictly factual.  And as dry as possible (fond thoughts of France, not the inclusion of vermouth, as it were).

Closes #165.

### DIFF
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -59,9 +59,8 @@ informative:
 
 Methods are defined for associating usage preferences
 with content that is obtained using the HTTP protocol.
-Two complementary methods are defined
-using either the Robots Exclusion Protocol
-or HTTP header fields.
+This document defines attachment methods
+using the Robots Exclusion Protocol and HTTP header fields.
 
 This document updates RFC 9309
 to allow for the inclusion of usage preferences.
--- a/draft-ietf-aipref-attach.md
+++ b/draft-ietf-aipref-attach.md
@@ -57,11 +57,11 @@ informative:
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-attach.md
```diff
@@ -57,11 +57,11 @@ informative:
...
-Content creators and other stakeholders might wish to signal
-their preferences about how their content
-might be consumed by automated systems.
-This document defines how preferences can be signaled
-as part of the acquisition of content in HTTP.
+Methods are defined for associating usage preferences
+with content that is obtained using the HTTP protocol.
+Two complementary methods are defined
+using either the Robots Exclusion Protocol
+or HTTP header fields.
```
ekr-ietf:
```suggestion
This document defines attachment methods
using the Robots Exclusion Protocol and HTTP header fields.
```

PR #225, opened 2026-08-20
Add Selection Step, Inference User, Inference System

tymrtn:
This refines #216 by defining the boundary between user-directed and system-directed inference. Nate's original commits are preserved at the base; this rebases onto -07 and adds two things.

First, a "selection step" that makes the user/system distinction mechanical: an operation by any party or process other than the user that identifies assets in response to a request (searching, browsing, ranking, recommending, resolving a description). Retrieval from among assets the user previously provided is not a selection step.

Second, the two category definitions rewritten around that term. User-directed: the user provided the asset or a direct reference the system fetches without any selection step. System-directed: everything else, including further assets fetched of the system's own choosing. Applies at time of use regardless of acquisition.

Under §6.4, categories not in the base document degrade to "unknown" permanently for conforming recipients. If inference categories are going in, they need to be here before this ships.

A decision-tree tool that walks scenarios through the vocabulary is at 
[The Inference Line](https://claude.ai/code/artifact/a4a1f206-1759-454b-8b03-90861d8d110b)

Addresses discussion in #216 from @natehake, @paul2keller, @ekr, @martinthomson, @TimidRobot, @chrisn, @lrosenthol, and @posth.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -281,7 +281,7 @@ Using an asset during inference where the human user, on whose behalf the infere
 
 ## Inference Directed by System {#ai-inference-system}
 
-Using an asset during inference in any other circumstance, including where the asset is identified through one or more selection steps.
+Using an asset during inference in any circumstance other than user direction, including where the asset is identified through one or more selection steps.
 Where a system begins from a user-provided reference and retrieves further assets of its own choosing, those further assets are selected by the system.
 
 This category applies at the time of use, regardless of when or how the asset was acquired.
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -106,6 +106,9 @@ Asset:
 Declaring party:
 : The entity that expresses a preference with regards to an Asset.
 
+Inference:
+: All use of an AI system, except use that modifies the learned parameters of an AI model.
+
 
 # Statements of Preference {#model}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -268,6 +274,19 @@ are used exclusively
...
+
+Using an asset during inference where the human user on whose behalf the inference is performed has provided the asset itself, or a reference to the asset that the system retrieves directly without performing any selection step.
+
+
+## Inference Directed by System {#ai-inference-system}
+
+Using an asset during inference in any other circumstance, including where the asset is identified through one or more selection steps.
+Where a system begins from a user-provided reference and retrieves further assets of its own choosing, those further assets are selected by the system.
+
+This category applies at the time of use, regardless of when or how the asset was acquired.
```
martinthomson, line 287: We've agreed not to have overlap between categories, but this creates an overlap with the infer-user category.

It also potentially overlaps with search, which is a problem with @natehake's original as well.  A specific exclusion, in one or other direction, could address that.  That is, you could say that if search applies, then either inference category does not apply.  I'd make a change to the search definition in that case.

(I don't think that the opposite makes a lot of sense, because search is a more specific scope than training or inference, even with the qualifications here.)

tymrtn: Infer-system's definition opens with "any other circumstance" — other than infer-user. So it can't overlap infer-user; it's built as everything infer-user doesn't cover. The one place I can see it leaking is the last sentence, if "regardless of when or how the asset was acquired" reads as an independent scope grant rather than timing detail inside the complement:

> Whether this category applies is determined at the time of use, according to how the asset was identified for that use rather than when it was acquired.

On search, can you help me understand the overlap you're seeing? [4.2 is about snippets and links to assets](https://datatracker.ietf.org/doc/html/draft-ietf-aipref-vocab-07#section-4.2), not generating content, so it shouldn't clear the inference gate at all. datatracker.ietf.org/doc/html/draft-ietf-aipref-vocab-07#section-4.2 AI-generated search results overviews already route to inference by the summary exclusion in 4.2.

martinthomson: The general consensus is that search is always inference.

There's some contention, but search is sometimes training, which is why we explicitly allow training if it is exclusively for a search application.  The same would apply to inference.

I don't think that "any other" is sufficiently clear.

draft-ietf-aipref-vocab.md
```diff
@@ -268,6 +274,19 @@ are used exclusively
 in ways that meet the above conditions.
 
 
+## Inference Directed by User {#ai-inference-user}
+
+Using an asset during inference where the human user on whose behalf the inference is performed has provided the asset itself, or a reference to the asset that the system retrieves directly without performing any selection step.
```
```suggestion
Using an asset during inference where the human user, on whose behalf the inference is performed, has provided the asset itself, or a reference to the asset that the system retrieves directly without performing any selection step.
```

draft-ietf-aipref-vocab.md
```diff
@@ -268,6 +274,19 @@ are used exclusively
...
+
+
+## Inference Directed by System {#ai-inference-system}
+
+Using an asset during inference in any other circumstance, including where the asset is identified through one or more selection steps.
```
```suggestion
Using an asset during inference in any circumstance other than user direction, including where the asset is identified through one or more selection steps.
```

lrosenthol (approved):
LGTM - just some minor editorials

martinthomson:
This is a useful contribution, but the discussion at our last interim in London ended up taking a slightly different (and perhaps less complex) path.  I suggest that you look at that, when it is available, which will be very soon. If this is still relevant, which I suspect it might not be, we can discuss that improvement relative to the new baseline.

tymrtn:
Thank you @martinthomson. I'll review the notes from the London session.

PR #228, opened 2026-08-24
More direct search=y processing conditions

martinthomson:
For some reason, this is hard.  The intent is to say that you can do whatever in your search application, including training and using models, provided that you meet the two conditions regarding referencing and excerpts.

This change uses more words to try to be clearer about that.  Less use of "this" and "that".


martinthomson:
Rolled up into our active branch that contains all the changes from the recent London interim.

PR #230, opened 2026-08-24
Generative AI model definition

martinthomson:
This changes the AI model training definition to point to a new definition "generative AI model", which uses exactly the same language we had in the definition of AI model training.  This is just reorganized.


kelleyk:
This seems very reasonable to me!

martinthomson:
Rolled up into our active branch that contains all the changes from the recent London interim.

PR #232, opened 2026-08-24
Try to clean up the text around S 4.2 per WG discussion.

ekr-ietf:


ekr-ietf:
MT writes in a separate issue: 
"In displaying titles or excerpts from assets,
changes to improve accessibility -" @ekr-ietf to revise.

Farzaneh-Unicode:
This is an excerpt from Krishna et al draft on search which I think can clarify the opt-out opt-in options in search: https://datatracker.ietf.org/doc/draft-madhavan-aipref-displaybasedpref/  4.1.  Indexing and Retrieval

   The act of allowing or disallowing content collected by web crawlers
   from being indexed or retrieved for purposes of display.  Such
   preference mechanism can also be applied for cases where digital
   content is not accessible.  In existing implementations, access
   preferences are typically expressed via the NOINDEX statement set in
   HTTP header or meta tags.

4.2.  Display text

   The act of allowing or disallowing a reproduction of text content
   collected by a web crawler, except for the title if specified, from
   the whole or parts of the content to display portions of that
   content.  In existing implementations preference on which text can be
   used for caption are expressed via the NOSNIPPET statement set in
   http header, HTML meta tags, or HTML tags properties (data-
   nosnippet).

4.3.  Display text length

   The act of limiting the number of characters as a textual display
   from content collected by a web crawler.  In existing implementations
   quotation length preferences are expressed via the max-snippet
   statement set in http header or HTML robots meta tags.

4.4.  Exact text match

   The act of limiting text content to only an exact match if displaying
   text content from the document.  If this preference is present, text
   content must be quoted as is or use avoided and an explicit link back
   to the source of the document used in that instance.  One example of
   existing implementation of text quotation preferences is notranslate.



4.5.  Image preview

   The act of limiting usage and size of images.  In existing
   implementations image preview preferences are typically expressed via
   the max-image-preview statement set in http header or HTML meta tags.

4.6.  Video preview

   The act of limiting usage and length of videos.  In existing
   implementations video preview preferences are typically expressed via
   the max-video-preview statement set in http header or HTML robots
   meta tags.

draft-ietf-aipref-vocab.md
```diff
@@ -243,15 +243,10 @@ This category of use only applies under the following conditions:
...
-Translation, transcription, or text-to-speech
-are examples of non-substantive changes
-that could help users understand what is being presented.
-Where existing controls restrict presentation of these items,
-such as limitations on snippet size,
-those apply before any changes.
+Changes such as translation, transcription, or text-to-speech of the
+presentation of titles or excerpts from assets are included in this
+category. These changes must not involve the generation of new
+material, though such changes may be permissible for other reasons.
```
martinthomson, line 249:
```suggestion
In displaying titles or excerpts from assets,
changes to improve accessibility,
such as translation, transcription, or text-to-speech,
are included in this category.
These changes must not involve the generation of new material,
though such changes may be permissible for other reasons.
```

chrisn:
What is the intent of "though such changes may be permissible for other reasons? It seems a bit open-ended.

martinthomson:
Rolled up into a larger changeset, so closing this for now.

martinthomson:
I just wanted to flag this @chrisn as something you might want to raise on #250:

> What is the intent of "though such changes may be permissible for other reasons? It seems a bit open-ended.

My understanding is that this refers to the possibility that a `search=y` allowance could also be paired with `ai-use=y`, which would allow for far more than simple permutations.  Or it could refer to Section 5.2, which talks about things like agreements.

chrisn:
Thanks, I'll review in the context of #250.

PR #233, opened 2026-08-24
Attempt to directly address agreements

### LINKED ISSUE #192: Introduction clarity re: relation to existing laws
### Confirmation

- [x] I have reviewed our [existing issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen) and this issue is not present there.

### Issue Description

It is clear from the debate in this working group over the past year that there is significant confusion about what legal implications, if any, the work of this group is aiming to have on creators globally. 

For example, the text of the current vocab draft ([draft-ietf-aipref-vocab-05](https://www.ietf.org/archive/id/draft-ietf-aipref-vocab-05.html)) says that it “is intended to be used in jurisdictions where expressing preferences results in legal obligations … without prejudice to applicable laws …,” but it is completely unclear (and silent about) what jurisdictions, legal obligations, or laws that it is referring to, exactly. 

Obviously, this working group, as part of a technical standards body, is not trying to hijack the role of a legislature or regulator.  It is not appropriate for us to design standards that may affect legal protections that millions of creators across the world have with relation to their countries’ domestic copyright, contract, and other laws. 

To make it clear that this working group is not aiming to override copyright owners’ existing legal protections in their respective countries, or to override creators’ existing terms of service, I propose to clarify the language in the introductory section.

I think that clarity will also improve uptake of the standardized language, better fulfilling the objectives of this WG. 

This issue applies to the vocabulary document (including some sections later on other than the introduction) and we should pursue similar clarification for the attachment document, to be clear that nothing in that document has any legal effect either - but I'll leave that for future issue submissions. 

### Proposed Resolution

I propose starting by strengthening language in the introductory section.
 
It currently reads: 

>   The vocabulary is intended to be used in jurisdictions where
>    expressing preferences results in legal obligations, as well as where
>    there are no associated legal obligations.  In either case,
>    expressing preferences is without prejudice to applicable laws,
>    including the applicability of exceptions and limitations to
>    Copyright.

Proposed edit: 

> For clarity, expressing preferences is without prejudice to applicable laws, 
> including the applicability of exceptions and limitations to copyright. 
> In addition to statutory or common law which may apply, the expression of preferences 
> is not intended to overwrite the legal preferences or restrictions that may be embodied
>  in a website’s Terms of Use (or comparable page), and in the event of any conflict
>  between these preferences and a website’s Terms of Use, the Terms of Use shall prevail as a matter of law.

martinthomson:
This is a minimal implementation of what @maxgendler proposes in #192.  It adds it to Section 3.2, which is risky, but that has the advantage of making the list there 5 items rather than 4 (and 4 is a number best avoided as a number of list items).

Closes #192.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -194,6 +194,7 @@ specification should understand that it does not:
 
 - ensure that preferences are followed;
 - address if, how, or when preferences should be followed or not-followed;
+- override agreements between parties;
 - address technical, legal, contractual, or other mechanisms that might create
   a stronger requirement to follow or not follow preferences;
 - consider situations or purposes that might justify following or not-following


martinthomson:
Hmm, I think that my list construction is not correct.

PR #234, opened 2026-08-25
Replace "asset" term with "resource"

kelleyk:
This morning, I was thinking about the concerns that were raised yesterday about the use of the term "asset".   I think that someone had suggested replacing it with "artifact", which also doesn't seem exactly right.  What if we use "resource" instead?

I've also added RFC 3986 as an informative reference and noted in the Resource definition that this document uses a narrower sense of the term than RFC 3986's more general usage.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -40,6 +40,7 @@ normative:
 
 informative:
   UTF8: RFC3629
+  URI: RFC3986
   ATTACH:
     title: "A Vocabulary For Expressing AI Usage Preferences"
     date: draft-ietf-aipref-attach-date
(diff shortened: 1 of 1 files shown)


ekr-ietf:
I could live with"resource"

lrosenthol (approved):
I can live with this

chrisn:
This is OK for me too. I previously raised a [concern](https://github.com/ietf-wg-aipref/drafts/issues/163#issuecomment-4247065161) about "resource", so the the reference to RFC 3986 was important to emphasise it being a technical term - but not to suggest that the vocabulary only applies to resources fetched via HTTP.

martinthomson:
In HTTP, a resource is an abstract notion that does not have the sort of concrete representation that this uses.  Why not "artifact"?

martinthomson:
Based on discussion about naming that we had at the last interim, I'm going to close this.  The request was that we retain "asset" for now, to avoid terminology confusion while the document is churning.

PR #236, opened 2026-08-25
Restore "make available for use" in generative defintion

martinthomson:
An improvement on #230 to include @kelleyk's original phrasing that covered open weight models.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -107,7 +107,7 @@ Declaring party:
 : The entity that expresses a preference with regards to an Asset.
 
 Generative AI model:
-: An AI model that is used or made available for use
+: An AI model that is used, or made available for use,
   in generating synthetic content
   in one or more modalities.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -106,6 +106,11 @@ Asset:
 Declaring party:
 : The entity that expresses a preference with regards to an Asset.
 
+Generative AI model:
+: An AI model that is used or made available for use
```
kelleyk: Can we put commas around this, so that it's "used, or made available for use, ..."?  I think that that reads a little bit more easily.

martinthomson:
```suggestion
: An AI model that is used, or made available for use,
```

lrosenthol (approved):

chrisn (approved):

martinthomson:
We took on a version of this into the working branch that was discussed at the last interim. The pull request for that change will supersede this.

Issue #226, opened 2026-08-24
Examplar vocab section contains material that is general purpose


lrosenthol:
### Confirmation

- [x] I have reviewed our [existing issues](https://github.com/ietf-wg-aipref/drafts/issues?q=state%3Aopen) and this issue is not present there.
- [x] I have read and understand the [IETF NOTE WELL](https://www.ietf.org/about/note-well/).

### Issue Description

some of the subsections of the "Exemplary Serialization Format" section should be in their own top level sections (or under some other general purpose section)

- Extensions.  Since this concept is called out/referenced earlier in the document in a general purpose statement ("Recipients that implement this specification will understand the vocabulary terms defined in {{vocab}}, but they might not understand extensions; see {{extension}}. "), then we need to discuss this concept w/o any implication to serialization.

- alternative formats.  The text reads, "Any alternative format needs to define the mapping both from that format to the model used in this document and from the model to the alternative format. This includes any potential for extensions ({{extension}})." This has nothing to do with the exemplary format.

### Proposed Resolution

Move those two sections to their own top level sections

PR #237, opened 2026-08-25
Fix extensions

lrosenthol:
Addresses #226 by:
- renaming the extension labels and using the correct one
- up-levelling the alternate format section


lrosenthol:
Wrong comparison!

PR #238, opened 2026-08-25
Correct linking to extensions & alternate formats

lrosenthol:
Address #226 by changing the labels and up-levelling the alternative formats section

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -300,8 +300,9 @@ and AI Use ({{ai-use}}).
 
 ## Vocabulary Extensions {#vocab-extensions}
 
-Extensions to this vocabulary are defined
+Extensions to this vocabulary may be defined
 in a standards-track RFC that updates this document.
+Or they may be defined by other organizations.
 
 The definition of the extension MUST define
 how any potential overlap between usage categories is resolved.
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -298,10 +298,11 @@ including AI Training ({{train-ai}})
 and AI Use ({{ai-use}}).
 
 
-## Vocabulary Extensions {#vocab-extension}
+## Vocabulary Extensions {#vocab-extensions}
 
-Extensions to this vocabulary are defined
+Extensions to this vocabulary may be defined
 in a standards-track RFC that updates this document.
+Or they may be defined by other organizations.
```
martinthomson, line 305: This cut is one that we are still discussing, so I'd rather not take this.  The rest of this is useful though.

martinthomson (changes requested):

martinthomson:
I've rolled these changes into the working branch that came out of the discussions at our last interim.  I have not started to define the extension model, which will not be included in that branch.

draft-ietf-aipref-vocab.md
lrosenthol: Agreed - no problem holding this for the extensions work!

PR #242, opened 2026-08-26
Align the vocab term name and keyword to its definition and meaning

bquinn:
This also leaves open the possibility for other AI training terms to be added in the future without causing confusion


martinthomson:
This sort of renaming will have to be done relative to a new baseline, so I'm closing this for now.  (It's possible that the generative qualifier is not something we need to include in the names of things and maybe even less necessary in the labels. The understanding of AI we collectively share is rapidly converging on those things that are generative.)

PR #241, opened 2026-08-26
Distribution implies communicating preferences

martinthomson:
This point is important as a clarification of what we have around the separation of training and use.

Closes #239 as an alternative.

PR #239, opened 2026-08-26
Add model context documentation requirement

TimidRobot:
Closes #245 

Add model context documentation requirement

The goal of this PR is to close an interpretation/implicit loophole to;
- Ensure good actors are supported in complying with the AI Preferences
- Increase confidence of AI Preference declarers that their preferences will be followed by good actors

PR #243, opened 2026-08-26
Conveying training preferences with models

martinthomson:
Yet another alternative to this "make preferences known with the model" downstream.

Closes #239.
Closes #241.


martinthomson:
Rolled up into a larger changeset with updates from all the discussion we've had recently.

PR #244, opened 2026-08-26
Communicating preferences with models

martinthomson:
I don't think that this is quite right.  But some people thought it was worthwhile.

### DIFF
--- a/draft-ietf-aipref-vocab.md
+++ b/draft-ietf-aipref-vocab.md
@@ -107,9 +107,11 @@ Declaring party:
 : The entity that expresses a preference with regards to an Asset.
 
 Generative AI model:
-: An AI model that is used,
-  or made available for use,
+: An AI model that is used
   to generate synthetic content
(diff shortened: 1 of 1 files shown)


draft-ietf-aipref-vocab.md
```diff
@@ -106,6 +106,17 @@ Asset:
...
 : The entity that expresses a preference with regards to an Asset.
 
+Generative AI model:
+: An AI model that is used
+  to generate synthetic content
+  in one or more modalities.
+  Alternatively, an AI model that is made available for use
+  in generating synthetic content
+  in one or more modalities.
+  \[This does not include classification, ranking, or scoring,
```
bquinn, line 116: Alternative: "This definition does not include models used for classification, ranking or scoring, even if those models output minimal generative content such as explanations of the outputs of those operations"

(this doesn't answer Glenn's objection, but I believe it clarifies the meaning of this sentence)

PR #248, opened 2026-08-26
Explain what might be needed to document conformance

martinthomson:
This is in response to concerns raised by Elaine about the lack of documentation about what it means to "conform" to this specification. This doesn't change that, but aims to make it clear how people can communicate about their conformance more generally.


martinthomson:
Also rolled up into the larger changeset.

